PR #13, opened 2022-08-02
Fix alignment issues

### LINKED ISSUE #11: Alignment of argument buffers not guaranteed
**Describe the bug**
The buffer used to read and write arguments for calls needs to be aligned to 8 bytes, but since it is a `static mut [u8; N]`, no such alignment is guaranteed.

**To Reproduce**
Run the tests without --release

**Additional context**
Creating an 8-aligned byte buffer is not as trivial as you would expect.

krl:
fixes #11


miloszm (commented):
Please do not add Cargo.lock again :)

miloszm (changes requested):
Other than technicalities like 1 byte file arg.rs and the added Cargo.lock, I could approve

miloszm (approved):

PR #14, opened 2022-08-02
Make `self_id` available to contracts

ureeves:
Introduce a new buffer, of size statically checked by `dallo`, which will contain the `self_id` of the contract. This buffer is written to on deployment of the contract.


krl (changes requested):
I am all in favor of the SELF_ID being passed this way, however i do not think the caller implementation needs this extra overhead, and it could be passed through the arg_buffer.

Note that cryptographic hashing of the module memory will have to exclude the arg_buf, and adding one extra precaution is also extra headaches down the line.

ureeves:
@krl  and I decided to remove half of the changes. Specifically the changes related to `caller` need more fleshing out, and the introduction of a call stack might have undesired consequences for the future.

hatchery/src/instance.rs
```diff
@@ -48,6 +51,7 @@ impl Instance {
             arg_buf_ofs,
             arg_buf_len,
             heap_base,
+            self_id_buf_ofs,
```
krl, line 54: this is not really a "buffer", `self_id_ofs` is more appropriate

modules/box/src/lib.rs
```diff
@@ -29,6 +29,9 @@ static mut A: [u64; ARGBUF_LEN / 8] = [0; ARGBUF_LEN / 8];
 #[no_mangle]
 static AL: i32 = ARGBUF_LEN as i32;
 
+#[no_mangle]
+static SELF_ID: [u8; MODULE_ID_BYTES] = [0u8; MODULE_ID_BYTES];
```
krl, line 33: I would make this a SELF_ID: ModuleId rather than an opaque buffer, is there something speaking against this?

modules/callcenter/src/lib.rs
```diff
@@ -33,6 +36,10 @@ impl Callcenter {
     pub fn increment_counter(self: &mut State<Self>, counter_id: ModuleId) {
         self.transact(counter_id, "increment", ())
     }
+
+    pub fn calling_self(self: &mut State<Self>, id: ModuleId) -> bool {
+        self_id() == &id
```
krl, line 41: also, i think in case of calling top-level functions, a `dallo::self_id()` will communicate more clearly what's going on.

krl (changes requested):
SELF_ID in contracts should be of type `ModuleId` and not an opaque buffer, otherwise this looks good!

hatchery/src/instance.rs
ureeves: Makes sense, will change

krl (approved):

PR #29, opened 2022-08-04
Add self_snapshot test, various refactorings and fixes

krl:


dallo/src/ops.rs
```diff
@@ -14,6 +14,6 @@ mod ext {
     }
 }
 
-pub fn self_id() -> &'static ModuleId {
-    unsafe { &ext::SELF_ID }
+pub fn self_id() -> ModuleId {
```
ureeves, line 17: I'm not sure about this. Why would we copy `self_id` onto the contract's stack?

dallo/src/types.rs
```diff
@@ -0,0 +1,48 @@
...
+    Eq,
+    Archive,
+    Serialize,
+    Deserialize,
+    PartialOrd,
+    Ord,
+    Clone,
+    Copy,
+)]
+#[repr(C)]
```
ureeves, line 24: Yay FFI safety

modules/self_snapshot/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+    "Kristoffer Ström <kristoffer@dusk.network>",
+    "Miłosz Muszyński <milosz@dusk.network>"
+]
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+dallo = { path = "../../dallo", default-features = false }
```
ureeves, line 14: Nit: newline

modules/self_snapshot/Makefile
```diff
@@ -0,0 +1,6 @@
+all: ## Generate the optimized WASM for the contract given
+	@cargo rustc \
+		--manifest-path=./Cargo.toml \
+		--release \
+		--target wasm32-unknown-unknown \
+		-- -C link-args=-s
```
ureeves, line 6: There should be no need for this file. The global makefile takes care of this

modules/self_snapshot/rustfmt.toml
```diff
@@ -0,0 +1,2 @@
+max_width = 80
+wrap_comments = true
```
ureeves, line 2: Same with this file

ureeves (changes requested):
Looks good! Amazing that it already works.
There's some files that don't need to be there, and a decision - specifically making `dallo::self_id` return a copy of the static `ModuleId` - that I'd like to see justified.

dallo/src/ops.rs
krl: Resolving as per our out of band discussion. (The compiler will most likely remove any unnecessary copy anyway)

modules/self_snapshot/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+        //     "setting crossover from {:?} to {:?}",
+        //     self.crossover,
+        //     to
+        // );
+        self.crossover = to;
+        old_val
+    }
+
+    // updates crossover and returns the old value
+    pub fn self_call_test_a(self: &mut State<Self>, update: i32) -> i32 {
```
miloszm, line 53: I'm confused by the `self` here, why does it have a type, if so, that means, it is not a real self? Maybe naming it differently would be less confusing?

ureeves (approved):
LGTM

modules/self_snapshot/src/lib.rs
krl: The self type is for tying the lifetime of the self reference to the wrapper struct `State`, to leverage the borrow checker for self-calling. 

In essence it guarantees that when  transacting on `dallo::self_id()` no other aliases or borrows of `SELF` exist.

krl: This keeps the call structure consistent, since a `&mut self` is freely interchangable if no calling is needed for the execution.

miloszm (approved):
LGTM, one comment but it's cosmetic

PR #31, opened 2022-08-05
Add `caller` host function

### LINKED ISSUE #18: Introduce `caller` as a host call
There should be a host function allowing contracts to query who is calling them. Since it is possible the contract under execution is the first in the call stack, we should use the "zero module id" a signal for this.

Contracts should be able to determine under what conditions they execute their own code.

N/A

**Additional context**
The transfer contract - one of two genesis contracts - currently requires this functionality to assert that the `mint` call may only be called from the stake contract.

ureeves:
Resolves #18

### DIFF
--- a/hatchery/src/world.rs
+++ b/hatchery/src/world.rs
@@ -5,6 +5,8 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 mod event;
+mod stack;
+
 pub use event::{Event, Receipt};
 
 use std::cell::UnsafeCell;
@@ -17,6 +19,7 @@ use std::sync::Arc;
 use dallo::{ModuleId, Ser, MODULE_ID_BYTES};
 use parking_lot::ReentrantMutex;
 use rkyv::{archived_value, Archive, Deserialize, Infallible, Serialize};
+use stack::CallStack;
 use tempfile::tempdir;
 use wasmer::{imports, Exports, Function, Val};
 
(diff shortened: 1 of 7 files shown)


hatchery/src/world.rs
```diff
@@ -32,6 +35,7 @@ pub struct WorldInner {
     environments: BTreeMap<ModuleId, Env>,
     storage_path: PathBuf,
     events: Vec<Event>,
+    call_stack: Option<CallStack>,
```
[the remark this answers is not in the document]
krl: this would also make all the expects unneccesary

hatchery/src/world.rs
```diff
@@ -269,6 +302,32 @@ impl World {
...
+        let guard = self.0.lock();
+        let w = unsafe { &*guard.get() };
+
+        let caller = w
+            .call_stack
+            .as_ref()
+            .expect("call stack should be set")
+            .caller();
+
+        instance.with_arg_buffer(move |buf| {
```
krl, line 315: instance has a `let arg_ofs = self.write_to_arg_buffer(arg)?;` that you should use instead of manually writing memory

modules/callcenter/src/lib.rs
```diff
@@ -41,6 +41,16 @@ impl Callcenter {
     pub fn calling_self(self: &mut State<Self>, id: ModuleId) -> bool {
         dallo::self_id() == id
     }
+
+    pub fn call_self(self: &mut State<Self>) -> bool {
+        let self_id = dallo::self_id();
+        let caller = self.caller();
+
+        match self.caller().is_uninitialized() {
```
krl, line 49: this is neat!

krl (changes requested):

krl (approved):

Issue #26, opened 2022-08-04
Persisting the world state should also persist the currently deployed modules.


krl:
**Describe what you want implemented** 
When restoring a persisted world state, the modules deployed in that world should be available.

**Describe "Why" this is needed**
Re-deploying all modules again after restore is not a realistic solution.

ureeves:
Closed by https://github.com/dusk-network/hatchery/pull/32

PR #32, opened 2022-08-05
World persist and restore

miloszm:
World persist and restore functionality.
State of all modules can be persisted and then restored, without redeployment of modules.
Implemented mechanism is verified by test `world_persist_restore` contained in file `boxen.rs`.
This PR solves issue #26.

### DIFF
--- a/hatchery/src/instance.rs
+++ b/hatchery/src/instance.rs
@@ -208,7 +208,7 @@ impl Instance {
         self.id
     }
 
-    pub fn set_snapshot_id(&mut self, snapshot_id: SnapshotId) {
+    pub(crate) fn set_snapshot_id(&mut self, snapshot_id: SnapshotId) {
         self.snapshot_id = Some(snapshot_id);
     }
     pub fn snapshot_id(&self) -> Option<&SnapshotId> {
--- a/hatchery/src/instance.rs
+++ b/hatchery/src/instance.rs
@@ -17,8 +17,8 @@ use wasmer::NativeFunc;
 
 use crate::error::*;
 use crate::memory::MemHandler;
-use crate::world::World;
 use crate::snapshot::SnapshotId;
+use crate::world::World;
 
 #[derive(Debug)]
 pub struct Instance {
@@ -208,8 +208,7 @@ impl Instance {
         self.id
     }
 
-    pub fn set_snapshot_id(&mut self, snapshot_id: SnapshotId)
-    {
+    pub fn set_snapshot_id(&mut self, snapshot_id: SnapshotId) {
         self.snapshot_id = Some(snapshot_id);
     }
     pub fn snapshot_id(&self) -> Option<&SnapshotId> {
--- a/hatchery/src/instance.rs
+++ b/hatchery/src/instance.rs
@@ -18,6 +18,7 @@ use wasmer::NativeFunc;
 use crate::error::*;
 use crate::memory::MemHandler;
 use crate::world::World;
+use crate::snapshot::SnapshotId;
 
 #[derive(Debug)]
 pub struct Instance {
@@ -29,6 +30,7 @@ pub struct Instance {
     arg_buf_len: i32,
     heap_base: i32,
     self_id_ofs: i32,
+    snapshot_id: Option<SnapshotId>,
 }
 
 impl Instance {
@@ -52,6 +54,7 @@ impl Instance {
             arg_buf_len,
             heap_base,
             self_id_ofs,
+            snapshot_id: None,
         }
     }
 
@@ -205,6 +208,13 @@ impl Instance {
         self.id
     }
 
+    pub fn set_snapshot_id(&mut self, snapshot_id: SnapshotId)
+    {
+        self.snapshot_id = Some(snapshot_id);
+    }
+    pub fn snapshot_id(&self) -> Option<&SnapshotId> {
+        self.snapshot_id.as_ref()
+    }
     pub(crate) fn world(&self) -> &World {
         &self.world
     }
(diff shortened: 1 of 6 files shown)


hatchery/src/snapshot.rs
```diff
@@ -0,0 +1,138 @@
...
+            .map_err(PersistenceError)?;
+        let metadata = std::fs::metadata(self.path().as_path())
+            .map_err(PersistenceError)?;
+        let mut buffer = vec![0; metadata.len() as usize];
+        f.read(buffer.as_mut_slice()).map_err(PersistenceError)?;
+        Ok(buffer)
+    }
+}
+
+pub struct MemoryPath {
```
ureeves, line 60: This structure can be removed if you implement `SnapshotLike` directly on `Path`

hatchery/src/snapshot.rs
```diff
@@ -0,0 +1,138 @@
...
+            snapshot_id_to_name(snapshot_id),
+        ));
+        Ok(Snapshot {
+            path,
+            id: snapshot_id,
+        })
+    }
+
+    /// Saves current snapshot as uncompressed file.
+    pub fn save_uncompressed(
```
krl, line 110: Maybe just call this `save`?

hatchery/src/snapshot.rs
```diff
@@ -0,0 +1,138 @@
...
+#[repr(C)]
+pub struct SnapshotId([u8; SNAPSHOT_ID_BYTES]);
+impl SnapshotId {
+    pub const fn uninitialized() -> Self {
+        SnapshotId([0u8; SNAPSHOT_ID_BYTES])
+    }
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.0
+    }
+}
```
ureeves, line 39: Are we thinking that we might pass a snapshot to a contract? If not, I would say it doesn't need to be `repr(C)` and `uninitialized` might not make sense.

hatchery/src/world.rs
```diff
@@ -76,13 +77,47 @@ impl World {
         )))))
     }
 
+    pub fn persist(&self) -> Result<(), Error> {
```
krl, line 80: This should probably also return a SnapshotId?

krl (changes requested):

hatchery/src/snapshot.rs
```diff
@@ -0,0 +1,138 @@
...
+
+pub struct Snapshot {
+    path: PathBuf,
+    id: SnapshotId,
+}
+
+impl Snapshot {
+    pub fn new(memory_path: &MemoryPath) -> Result<Self, Error> {
+        let snapshot_id: SnapshotId = SnapshotId::from(
+            *blake3::hash(memory_path.read()?.as_slice()).as_bytes(),
```
ureeves, line 86: Shouldn't the snapshot id be somehow computed from the state itself and not just the path?

ureeves (changes requested):
A few requests and questions about `SnaphotId`s

hatchery/src/snapshot.rs
krl: Yes, ideally the snapshot state should be a combination of all instance ids, which themselves should be the hash of their memory excluding the argument buffer

hatchery/src/snapshot.rs
miloszm: we will need to store it, but point taken

hatchery/src/snapshot.rs
miloszm: snapshot is computed from a file's contents the path is pointing, see the little "read()?" there? ;)

hatchery/src/snapshot.rs
miloszm: But I wanted to have MemoryPath as a type, to pass it safely around and distinguish it from Path.
You suggest to do `type MemoryPath = Path` instead?

hatchery/src/snapshot.rs
miloszm: ok

hatchery/src/world.rs
miloszm: sounds good

hatchery/src/snapshot.rs
ureeves: If you want to keep a type alias, sure. It seems that you could just distinguish it by name though.

hatchery/src/snapshot.rs
ureeves: Missed the read indeed. My bad

hatchery/src/world.rs
miloszm: issue is that there are multiple snapshots, one per each module, we don't have global snapshot (yet), so I can't return snapshot id in this case

hatchery/src/snapshot.rs
miloszm: Issue is that it is very easy to make mistake, MemoryPath is something that is a current live memory for a module, to be distinguished from just passive snapshot. To avoid confusion in parameters, I wanted to keep this type really separate.

miloszm: This is called 'newtype' pattern - wrapping a core type within a single field struct (or a tuple)

ureeves (approved):
LGTM

hatchery/src/instance.rs
```diff
@@ -205,6 +208,12 @@ impl Instance {
         self.id
     }
 
+    pub fn set_snapshot_id(&mut self, snapshot_id: SnapshotId) {
```
krl, line 211: Is there a reason for this to be public? Seems like internal-only api to me. consider `pub(crate)` if neccesary

krl (changes requested):

krl (approved):

PR #34, opened 2022-08-08
Add `height` host call

### LINKED ISSUE #23: Allow module to query the block height
Modules should have the block height available for query.

This is necessary for modules to be able to express logic conditional to (blockchain) time passed.

N/A

**Additional context**
The stake contract currently has a "cooldown" period between a stake and the possibility to partake in the consensus.

ureeves:
Resolves #23

### DIFF
--- a/hatchery/tests/callcenter.rs
+++ b/hatchery/tests/callcenter.rs
@@ -12,14 +12,14 @@ pub fn world_center_counter_read() -> Result<(), Error> {
 
     let counter_id = world.deploy(module_bytecode!("counter"))?;
 
-    let value: Receipt<i64> = world.query(counter_id, "read_value", ())?;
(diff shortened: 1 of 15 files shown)


hatchery/tests/callcenter.rs
```diff
@@ -12,14 +12,14 @@ pub fn world_center_counter_read() -> Result<(), Error> {
 
     let counter_id = world.deploy(module_bytecode!("counter"))?;
 
-    let value: Receipt<i64> = world.query(counter_id, "read_value", ())?;
+    let value: Receipt<i64> = world.query(0, counter_id, "read_value", ())?;
```
krl, line 15: I'm not to happy with this being handled on every single query. We should instead make it a method on world.

krl (changes requested):

krl (approved):

PR #36, opened 2022-08-10
Add module metering

### LINKED ISSUE #19: Implement metering of modules
The execution of modules should be metered. This can be done by using the facilities `wasmer` provides.

This is required to allow for charging for the execution of a smart contract.

Previously, metering was done using by annotating the code with the costs of each section. This was abandoned in favor of the more elegant approach used in `wasmer`.  

**Additional context**
This implementation should allow the user to call the virtual machine, specifying the amount of gas they're willing to spend. If the execution runs out, it should return an (OutOfGas) error, otherwise it should return the actual gas spent.

ureeves:
Resolves #19

### DIFF
--- a/hatchery/src/error.rs
+++ b/hatchery/src/error.rs
@@ -4,10 +4,12 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+use std::io;
+
+use dallo::ModuleId;
 use rkyv::ser::serializers::{
     BufferSerializerError, CompositeSerializerError, FixedSizeScratchError,
 };
-use std::io;
 
 pub type Compo = CompositeSerializerError<
     BufferSerializerError,
@@ -25,7 +27,7 @@ pub enum Error {
     MissingModuleExport,
     CompositeSerializerError(Compo),
     PersistenceError(io::Error),
-    OutOfPoints,
+    OutOfPoints(ModuleId),
 }
 
 impl From<wasmer::InstantiationError> for Error {
--- a/hatchery/src/error.rs
+++ b/hatchery/src/error.rs
@@ -25,6 +25,7 @@ pub enum Error {
     MissingModuleExport,
     CompositeSerializerError(Compo),
     PersistenceError(io::Error),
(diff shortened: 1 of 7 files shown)


hatchery/src/error.rs
```diff
@@ -25,6 +27,7 @@ pub enum Error {
     MissingModuleExport,
     CompositeSerializerError(Compo),
     PersistenceError(io::Error),
+    OutOfPoints(ModuleId),
```
krl, line 30: I don't see the reason to include the ModuleId here

hatchery/src/world.rs
```diff
@@ -31,13 +33,17 @@ use crate::snapshot::{MemoryPath, Snapshot, SnapshotLike};
...
+const POINT_PASS_PERCENTAGE: u64 = 93;
+
 #[derive(Debug)]
 pub struct WorldInner {
     environments: BTreeMap<ModuleId, Env>,
     storage_path: PathBuf,
     events: Vec<Event>,
     call_stack: CallStack,
     height: u64,
+    limit: u64,
```
krl, line 46: The point limit does not belong to the world.

krl (changes requested):

hatchery/src/error.rs
ureeves: The user might appreciate knowing which contract was given too little points to operate.

krl: This can already be more elegantly deducted from what's on top of the call stack

ureeves: The call stack is currently not exposed to the caller

krl: it feels off to me, but not a big deal.

hatchery/src/world.rs
ureeves: Lets resolve this in favor of https://github.com/dusk-network/hatchery/issues/41

krl: postponed for later refactor.

krl (approved):

PR #42, opened 2022-08-12
Use bytecheck to deserialize safely

### LINKED ISSUE #35: Use `bytecheck` crate for `world.query` and `world.transact`
At the moment, it's possible to segfault the vm by providing an erroneous return type. In order to fix this, we should use the `bytecheck` crate for the return value when deserializing.

It should not be possible to crash the vm without unsafe.

krl:
fixes #35

### DIFF
--- a/dallo/src/helpers.rs
+++ b/dallo/src/helpers.rs
@@ -4,18 +4,14 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-use crate::SCRATCH_BUF_BYTES;
-
+use crate::{StandardDeserialize, SCRATCH_BUF_BYTES};
 use rkyv::ser::serializers::{
     BufferScratch, BufferSerializer, CompositeSerializer,
 };
 use rkyv::ser::Serializer;
-use rkyv::{archived_root, Archive, Deserialize, Serialize};
+use rkyv::{check_archived_root, Archive, Deserialize, Serialize};
 
-pub type StandardBufSerializer<'a> = CompositeSerializer<
-    BufferSerializer<&'a mut [u8]>,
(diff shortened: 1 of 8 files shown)


dallo/src/helpers.rs
```diff
@@ -4,31 +4,27 @@
...
 
-use crate::SCRATCH_BUF_BYTES;
-
+use crate::{StandardDeserialize, SCRATCH_BUF_BYTES};
 use rkyv::ser::serializers::{
     BufferScratch, BufferSerializer, CompositeSerializer,
 };
 use rkyv::ser::Serializer;
-use rkyv::{archived_root, Archive, Deserialize, Serialize};
+use rkyv::{check_archived_root, Archive, Deserialize, Serialize};
```
ZER0, line 12: Unless there is a specific reason, it usually in Rust it's more idiomatic having `checked` / `unchecked` as suffix instead of prefix (see the standard library for example).

krl: This is the name in `rkyv`, not something we control.

dallo/src/types.rs
```diff
@@ -5,10 +5,33 @@
...
+
+pub trait StandardDeserialize<T>:
+    Deserialize<T, Infallible> + for<'a> CheckBytes<DefaultValidator<'a>>
+{
+}
+
+impl<T, U> StandardDeserialize<T> for U where
+    U: Deserialize<T, Infallible> + for<'a> CheckBytes<DefaultValidator<'a>>
+{
+}
```
ureeves, line 33: Consider not introducing this wrapper trait. The user will still have to just implement `Deserialize<_, Infallible>` and `CheckBytes<DefaultValidator<_>>`.

ureeves (approved):
Looks good. I would consider removing the wrapper trait below, but maybe that's just me.

PR #45, opened 2022-08-15
Move the argument buffer handlnig to `dallo`

krl:
Make `query`, `query_raw`, `caller`, `height` callable from a top-level function.


ureeves (approved):
Looks good. We'll have to revisit `self_snapshot`

PR #51, opened 2022-08-22
Add host module support

### LINKED ISSUE #49: Introduce the concept of a host module
Modules should have access to reserved queries through the `query` host call. These calls should be defined upstream the user using a trait named `HostModule`, and executed on the host .

Host modules should be registered with a `World`, and the trait should look similar to:

```rust
trait HostModule<Arg, Ret> {
    fn host_query(arg: Arg) -> Ret;
}
```

This is required to allow for the introduction of functions that would otherwise cost a module too many points to perform. These can include hashing, verifying zero-knowledge proofs, etc...

ureeves:
This introduces a new host import `nq`, allowing modules to execute arbitrary queries that are executed on the host and defined upstream via the `NativeQuery` trait.

Resolves #49

### DIFF
--- a/hatchery/src/world.rs
+++ b/hatchery/src/world.rs
@@ -5,10 +5,12 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 mod event;
+mod native;
 mod stack;
 mod store;
 
 pub use event::{Event, Receipt};
+pub use native::NativeQuery;
 
 use std::cell::UnsafeCell;
 use std::collections::BTreeMap;
@@ -20,6 +22,7 @@ use std::sync::Arc;
 use dallo::{
     ModuleId, StandardBufSerializer, StandardDeserialize, MODULE_ID_BYTES,
 };
+use native::NativeQueries;
 use parking_lot::ReentrantMutex;
 use rkyv::{Archive, Serialize};
 use stack::CallStack;
@@ -41,6 +44,7 @@ const POINT_PASS_PERCENTAGE: u64 = 93;
 #[derive(Debug)]
 pub struct WorldInner {
     environments: BTreeMap<ModuleId, Env>,
+    native_queries: NativeQueries,
     storage_path: PathBuf,
     events: Vec<Event>,
     call_stack: CallStack,
@@ -72,6 +76,7 @@ impl World {
     {
         World(Arc::new(ReentrantMutex::new(UnsafeCell::new(WorldInner {
             environments: BTreeMap::new(),
+            native_queries: NativeQueries::new(),
             storage_path: path.into(),
             events: vec![],
             call_stack: CallStack::default(),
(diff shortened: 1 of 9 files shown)


hatchery/src/world.rs
```diff
@@ -459,6 +489,30 @@ fn host_query(
...
+) -> u32 {
+    let name_adr = name_adr as usize;
+    let name_len = name_len as usize;
+
+    let instance = env.inner();
+
+    let name = instance.with_memory(|buf| {
+        let name = core::str::from_utf8(&buf[name_adr..][..name_len])
+            .expect("TODO, error out cleaner");
+        String::from(name)
```
miloszm, line 506: this could be `name.to_owned()` - conveying the intention bit better IMO

hatchery/src/world.rs
```diff
@@ -193,6 +200,17 @@ impl World {
         Ok(id)
     }
 
+    /// Deploys a [`HostQuery`] with the given module ID.
+    pub fn deploy_host<Q>(&mut self, name: &'static str, query: Q)
```
miloszm, line 204: name could (should) be `Cow<'static, str>`, this would let deploy_host accept any kind of string with no penalty

hatchery/src/world.rs
```diff
@@ -193,6 +200,17 @@ impl World {
         Ok(id)
     }
 
+    /// Deploys a [`HostQuery`] with the given module ID.
```
krl, line 203: given module id?

hatchery/src/world.rs
```diff
@@ -459,6 +487,31 @@ fn host_query(
         .expect("TODO: error handling")
 }
 
+fn host_host_query(
```
krl, line 490: not so sure about the host_host thing, is there not a clearer way of naming this?

miloszm (approved):

krl (commented):
Aside from naming, this looks promising.

I'll have a closer look as how the buffer handling is actually implemented.

hatchery/src/world.rs
ureeves: When I first started I wanted host queries to piggy back on the normal query infra. Forgot to update the comment. Well spotted.

hatchery/src/world.rs
ureeves: I don't like it much either. Maybe I can come up with another name besides "host".

hatchery/src/world/native.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    pub fn call(&self, name: &str, buf: &mut [u8], len: u32) -> Option<u32> {
+        self.map.get(name).map(|host_query| host_query(buf, len))
+    }
+}
+
+/// A query executable on the host.
+///
+/// The argument the module has placed in the buffer will be present from
+/// `0..u32`, and should be processed first. Once this is done, the implementor
```
miloszm, line 42: nitpick - I was staring at 0..u32 for 3 minutes trying to understand it :), it could be sth like 0..value_of(arg:u32)..

miloszm (approved):
1 nit remark, otherwise lgtm

krl (approved):

PR #56, opened 2022-08-26
Remove validation guest-side, debloating the module sizes

krl:


modules/box/Cargo.toml
```diff
@@ -5,11 +5,12 @@ authors = [
...
-
 license = "MPL-2.0"
 
 [dependencies]
 dallo = { path = "../../dallo", default-features = false }
 
 [lib]
 crate-type = ["cdylib", "rlib"]
+
+[workspace]
```
ureeves, line 16: And these should be removed

ureeves (changes requested):
The validation changes LGTM, but the Makefile and workspace changes increase complexity quite a bit and slow compilation times.
Here's PR https://github.com/dusk-network/hatchery/pull/58 addressing this. What do you think?

ureeves (approved):
LGTM

PR #60, opened 2022-09-02
Serialized modules

krl:
Serialization of modules


vmx/src/.#error.rs
```diff
@@ -1 +0,0 @@
-krl@x.1050905:1662051485
```
ureeves, line 1: ?

ureeves (approved):
Cool!

vmx/src/.#error.rs
krl: emacs litter, apologies

PR #65, opened 2022-09-19
Callcenter

krl:


modules/callcenter/src/lib.rs
```diff
@@ -26,11 +23,12 @@ static mut STATE: State<Callcenter> = State::new(Callcenter);
 
 impl Callcenter {
     pub fn query_counter(&self, counter_id: ModuleId) -> i64 {
-        dallo::query(counter_id, "read_value", ())
+        let result = uplink::query(counter_id, "read_value", ());
+        uplink::debug!("Quiring counter, got {}", result);
```
ureeves:
```suggestion
        uplink::debug!("Querying counter, got {}", result);
```

uplink/Cargo.toml
```diff
@@ -8,6 +8,7 @@ license = "MPL-2.0"
 [dependencies]
 rkyv = { version = "0.7", default-features = false, features = ["size_32", "alloc"] }
 bytecheck = { version = "0.6", default-features = false }
+wee_alloc = "0.4.5"
```
ureeves, line 11: Both are fine, but we should decide if we want the usage of `wee_alloc` to be enforced in `uplink`, or if we want to leave it to contract writers. Keep in mind that there's an incentive to keep the bytecode small anyway.

uplink/src/helpers.rs
```diff
@@ -25,12 +25,24 @@ where
     R: for<'a> Serialize<StandardBufSerializer<'a>>,
     F: Fn(A) -> R,
 {
+    crate::debug!("wrap query");
     with_arg_buf(|buf| {
         let slice = &buf[..arg_len as usize];
+
+        crate::debug!("slice {:?}", slice);
```
ureeves: Should we keep these debug statements?

uplink/src/state.rs
```diff
@@ -136,26 +125,28 @@ where
...
 
-    let name = raw.name();
-    let arg_len = raw.arg_bytes().len() as u32;
-    let ret_len = extern_query(mod_id, name, arg_len);
+    // let name = raw.name();
+    // let arg_len = raw.arg_bytes().len() as u32;
+    // let ret_len = unsafe { ext::q() };
 
-    with_arg_buf(|buf| RawResult::new(&buf[..ret_len as usize]))
+    // with_arg_buf(|buf| RawResult::new(&buf[..ret_len as usize]))
```
ureeves, line 140: ?

uplink/src/state.rs
```diff
@@ -166,13 +157,14 @@ where
...
-    with_arg_buf(|buf| {
-        let slice = &buf[..ret_len as usize];
-        let ret = unsafe { archived_root::<Ret>(slice) };
-        ret.deserialize(&mut Infallible).expect("Infallible")
-    })
+    // with_arg_buf(|buf| {
+    //     let slice = &buf[..ret_len as usize];
+    //     let ret = unsafe { archived_root::<Ret>(slice) };
+    //     ret.deserialize(&mut Infallible).expect("Infallible")
+    // })
```
ureeves, line 166: ?x2

uplink/src/types.rs
```diff
@@ -28,14 +27,15 @@ pub const MODULE_ID_BYTES: usize = 32;
...
+    CheckBytes,
     Deserialize,
     PartialOrd,
     Ord,
     Hash,
     Clone,
     Copy,
 )]
-#[archive_attr(derive(CheckBytes))]
+#[archive(as = "Self")]
```
ureeves, line 38: This is indeed far better than an `ArchivedModuleId`

miloszm (changes requested):
All looks good but make test fails, probably because produced wasm file is too big

miloszm (approved):

modules/micro/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+use uplink;
+
+static mut A: u32 = 42;
+
+#[no_mangle]
+unsafe fn change(to: u32) -> u32 {
+    let r = A;
+    A = to;
+    r
+}
```
ureeves, line 20: Does this work? Isn't it just a coincidence with the call convention?

ureeves (approved):
Small things here and there, but looks good.

PR #71, opened 2022-09-22
Make `wee_alloc` optional

### LINKED ISSUE #68: Do not mandate weealloc
At the moment weealloc is hardcoded in the uplink communication library, this is unneccesaily restrictive and it should be up to module autors to choose allocation strategy.

Flexibility in choosing allocators.

ureeves:
The global allocator can either be set by the user, or by the uplink when selecting the `wee_alloc` feature.

Resolves #68

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ modules: ## Build WASM modules
 	 	          -o target/stripped/%
 
 test: modules assert-counter-module-small ## Run the tests
-	cargo test \
+	@cargo test \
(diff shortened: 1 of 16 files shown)


krl (changes requested):
I really think it should be optional as in you need to actively opt-in in the module itself, not as a feature.

ureeves:
The global allocator must now be set by the user of the uplink

krl (approved):

PR #74, opened 2022-09-23
Caller implementation

krl:
This completes the `Callcenter` test ports


ureeves (approved):
Looks good. I like the change in the import call convention.

miloszm (approved):

PR #77, opened 2022-09-27
Implement queries on the host

ureeves:
Host queries are included in by the virtual machine and are available to
all contracts equally. They can be registered with the VM and offer type
safety to the module writer, but not to the implementer.


vmx/src/vm/host.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    pub fn call(&self, name: &str, buf: &mut [u8], len: u32) -> Option<u32> {
+        self.map.get(name).map(|host_query| host_query(buf, len))
+    }
+}
+
+/// A query executable on the host.
+///
+/// The buffer containing the argument the module used to call the query
+/// together with its length are passed as arguments to the function, and should
```
krl, line 37: Is it not reduntant to use a buffer of length n and also an explicit length?

ureeves: I suppose it is. I'll make the change

ureeves: Actually, now that I think better about it, it isn't redundant since we're using this buffer to read as well as write. If we assume the length of the buffer is the length of the argument, we can't then write a return that is larger than said argument length.

vmx/src/vm.rs
```diff
@@ -97,6 +104,15 @@ impl VM {
         })
     }
 
+    /// Registers a [`HostQuery`] with the given `name`.
+    pub fn register_host_query<Q>(&mut self, name: &'static str, query: Q)
```
miloszm, line 108: Are we sure we always want to accept only static strings as query names? Otherwise Cow could be used for str.

vmx/src/vm.rs
```diff
@@ -4,6 +4,9 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+mod host;
```
krl, line 7: If you really feel it's neccesary to make a submodule, please use the convention of module_name/mod.rs to not have multiple directories levels of related files.

krl (changes requested):

vmx/src/vm.rs
ureeves: I'm a little puzzled with this. Since [rust 2018 edition](https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html) `mod.rs` files are no longer needed when making submodules. Do we really want to keep the old convention?
If you look into the `hatchery` on the `main` branch you'll see an example where you don't really see nesting either.

vmx/src/vm.rs
```diff
@@ -185,6 +201,16 @@ impl VM {
         session.query(id, method_name, arg)
     }
 
+    pub(crate) fn host_query(
```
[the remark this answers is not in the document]
ureeves: I feel like this should be the subject of an improvement in a different PR. As it stands these queries are as useful as they need to be. I do agree they could be more useful and this is a possible improvement.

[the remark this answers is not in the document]
krl: I really hate having directories and files named the same, is there another way where we don't?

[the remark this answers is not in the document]
ureeves: Yes. Lets skirt the issue for now and put it in the same file

krl (approved):

PR #78, opened 2022-09-28
Module memory commit and restore mechanism

miloszm:
Module memory commit and restore mechanism
Implements the following scenario:

state = s1
commit1 = commit()
state = s2
commit2 = commit()

restore(commit1)
assert_eq(state, s1)
restore(commit2)
assert_eq(state, s2)

### DIFF
--- a/vmx/tests/counter.rs
+++ b/vmx/tests/counter.rs
@@ -56,7 +56,7 @@ fn counter_read_write_session() -> Result<(), Error> {
 
     other_session.transact::<(), ()>(id, "increment", ())?;
 
-    let _commit_id = vm.commit(&id)?;
+    let _commit_id = other_session.commit(&id)?;
 
     // session committed, new value accessible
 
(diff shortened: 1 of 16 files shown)


vmx/tests/counter.rs
```diff
@@ -56,11 +56,52 @@ fn counter_read_write_session() -> Result<(), Error> {
...
+    let id = vm.deploy(module_bytecode!("counter"))?;
+
+    // commit 1
+    let mut session_1 = vm.session();
+
+    assert_eq!(session_1.query::<(), i64>(id, "read_value", ())?, 0xfc);
+
+    session_1.transact::<(), ()>(id, "increment", ())?;
+
+    let commit_1 = vm.commit(&id)?;
```
ureeves, line 80: I don't think commits should be at a module level, they should rather commit the result of a session. The user should have no control over which modules are persisted at all.

ureeves (changes requested):
The behavior of the commit mechanism should be probably be at `Session` level rather than at `VM` level. On another note, I tend to think that the concept of a session is rather useless currently, since there is no additional function from just using the `VM`.

miloszm:
has been superseded by vmx-session-commit-restore branch

PR #88, opened 2022-10-04
VM Session Commit-Restore Mechanism

### LINKED ISSUE #33: World persist and restore refinement
Persist snapshot ids a use diffed snapshots

Current implementation does not persist snapshot ids, but rather only the content. Also, diff'ed snapshots are not used.

We could use not diff'ed snapshots but they take up more space.

**Additional context**
N/A

miloszm:
VM Session Commit-Restore Mechanism

makes the following pseudocode test scenario work:

create VM
deploy modA and modB
assert state of modA is SA0, state of modB is SB0
start session S
set state of modA to SA1, state of modB to SB1
commit session S, note commit id as CID1
set state of modA to SA2, state of modB to SB2
assert state of modA is SA2, state of modB is SB2
restore session S with CID1
assert state of modA is SA1, state of modB is SB1

Closes #33


vmx/src/commit.rs
```diff
@@ -0,0 +1,167 @@
...
+impl SessionCommitId {
+    pub fn uninitialized() -> Self {
+        SessionCommitId([0; COMMIT_ID_BYTES])
+    }
+
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.0[..]
+    }
+
+    pub fn add(&mut self, module_commit_id: &ModuleCommitId) {
```
krl, line 67: This should not be public

vmx/src/memory_path.rs
```diff
@@ -0,0 +1,40 @@
...
+
+impl MemoryPath {
+    pub fn new<P: AsRef<Path>>(path: P) -> Self
+    where
+        P: Into<PathBuf>,
+    {
+        MemoryPath { path: path.into() }
+    }
+
+    pub fn read(&self) -> Result<Vec<u8>, Error> {
```
krl, line 25: I don't see the reason for reading the whole file into memory when we already have the memory maps.

vmx/src/session.rs
```diff
@@ -68,6 +38,7 @@ impl Session {
             memory_handler: MemoryHandler::new(vm.clone()),
             vm,
             callstack: Arc::new(RwLock::new(vec![])),
+            current_commit_id: None,
```
krl, line 41: Commit ids should be calculated at the end of the session, not kept in it.

vmx/src/session.rs
```diff
@@ -179,4 +150,49 @@ impl Session {
         let mut s = self.callstack.write();
         s.pop();
     }
+
+    pub fn commit(&mut self) -> Result<SessionCommitId, Error> {
```
krl, line 154: commit should take self by value

vmx/src/vm.rs
```diff
@@ -213,6 +146,80 @@ impl VM {
...
+                self.path_to_module_commit(module_id, &module_commit_id)
+            },
+        )
+    }
+
+    pub(crate) fn add_session_commit(&mut self, session_commit: SessionCommit) {
+        self.inner.write().session_commits.add(session_commit);
+    }
+
+    pub(crate) fn restore_session(
```
krl, line 195: This is a bit of a conceptual issue, i think it would be more reasonable to have a VM::from_commit_id(...); - the `SessionCommitId` is redundant and should be called `CommitId`

vmx/tests/box.rs
```diff
@@ -19,7 +19,7 @@ pub fn box_set_get() -> Result<(), Error> {
     let mut session = vm.session();
 
     session.transact::<i16, ()>(id, "set", 0x11)?;
-    session.commit(&id)?;
+    session.commit()?;
```
krl, line 22: Commit here should not be necessary, since the session has already mutated the box content.

vmx/tests/session_commit.rs
```diff
@@ -0,0 +1,135 @@
...
+    assert_eq!(session.query::<(), i64>(id_1, "read_value", ())?, 0xfd);
+    assert_eq!(
+        session.query::<_, Option<i16>>(id_2, "get", ())?,
+        Some(0x11)
+    );
+    Ok(())
+}
+
+#[test]
+fn multiple_commits_per_session() -> Result<(), Error> {
```
krl, line 110: this should not be a thing at all.

krl (changes requested):

vmx/src/commit.rs
miloszm: done

vmx/src/memory_path.rs
miloszm: done

vmx/tests/session_commit.rs
miloszm: done

vmx/tests/box.rs
miloszm: done

vmx/src/session.rs
miloszm: done

vmx/src/vm.rs
miloszm: I cannot create VM out of session id as it is VM who has the location of all session's persistence, the base location..

krl: ok let's keep it like this for now

krl (approved):

Issue #89, opened 2022-10-06
Session Commits Persistence


miloszm:
#### Summary
> 💡 After system is restarted, we need to be able to find and restore session's commits that have been performed in the past.

#### Possible solution design or implementation
> 💡 TBD

#### Additional context
> 💡 The following minimal test T1 expressed as pseudocode should work:

start VM
deploy module M1 with initial state S1
start session S
assert state of M1 is S1
set state of M1 to S2
commit session S, obtain commit id CID1
drop session S
start session S
restore session to CID1
assert state of M1 is S2

Note that this issue is about session level persistence:

Session level persistence - session's commit method produces a session_commit_id which can be used at later time as argument to session's restore method. Note, that here we can distinguish: a) persistence of particular commits, under the assumption that the user stores commit ids, as well as b) persistence of commit ids themselves, under the assumption that the user does not store them.

ureeves:
Closed by #98

PR #98, opened 2022-10-13
Session Commits Persistence Implementation

miloszm:
Session Commits Persistence Implementation

Solves issue #89


vmx/src/commit.rs
```diff
@@ -4,17 +4,40 @@
...
+    CheckBytes,
+    Deserialize,
+    PartialOrd,
+    Ord,
+    Hash,
+    Clone,
+    Copy,
+)]
+#[archive(as = "Self")]
+#[repr(C)]
```
ureeves, line 40: Is it necessary to make these `#[repr(C]`?

vmx/src/commit.rs
```diff
@@ -45,7 +68,21 @@ impl core::fmt::Debug for ModuleCommitId {
...
```
ureeves, line 85: Ditto

ureeves (approved):
LGTM, one nit

vmx/src/commit.rs
miloszm: this was done because ModuleId has it as well, and these objects have identical structure, we could argue this is not needed here but it might be needed as commit ids could be exported in the future

vmx/src/commit.rs
```diff
@@ -147,6 +197,46 @@ impl SessionCommits {
             None => Err(SessionError("unknown session commit id".into())),
         }
     }
+
+    pub fn read<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
```
krl, line 201: naming here, using "read/write" is very generic and overlaps with the Read and Write traits unnecessarily, consider store/restore or something similar (in all cases where read/write is used)

krl (approved):

PR #106, opened 2022-10-18
Rename to `piecrust` and `piecrust-uplink`

ureeves:
Renames `vmx` to `piecrust` and `uplink` to `piecrust-uplink`, and add a more comprehensive README.


.github/workflows/build.yml
```diff
@@ -16,6 +16,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
+          toolchain: nightly
```
miloszm, line 19: are we bringing back the nightly?

miloszm (changes requested):

krl (approved):

README.md
```diff
@@ -1,5 +1,35 @@
...
+```
+
+## Build and Test
+
+To build and test the crate one will need a
+[Rust](https://www.rust-lang.org/tools/install) toolchain, Make, and the
+`wasm-tools` binary.
+
+```sh
+sudo apt install -y make
```
krl, line 32: If we are debian/ubuntu specific we should at least mention that? I find it odd to assume.

.github/workflows/build.yml
ureeves: In this PR I change the toolchain file to be `rust-toolchain.toml`. This behaves differently and the action doesn't pick it up - see the action performed on the [penultimate commit](https://github.com/dusk-network/hatchery/actions/runs/3273676619). When running `make test` locally the correct toolchain will always be loaded.
We have to be careful in the CI just as before, but at least we will know how to solve the problem quickly.

miloszm (approved):

Issue #111, opened 2022-10-24
Move `deploy`ment of modules to `Session` level


ureeves:
#### Summary
Currently the deployment of modules happens at the level of the `VM` struct. This is undesirable since this means that a deployment is "global" - meaning for all session. All changes of `VM` state should be done via a `Session` and subsequent `commit`ment.

#### Possible solution design or implementation
Move the `deploy` and `deploy_with_id` functions to take `Session` as a receiver, as opposed to `VM`. Due to this change, the management of the state will need to move more towards a `Session` rather than `VM` as well.

#### Additional context
Since modifications of the VMs state happen should happen at the level of a `Session` this also means ideally that two `Session`s could be ran in parallel. To achieve this care needs to be taken with how data is structured on disk.

PR #112, opened 2022-10-25
Moved Deploy to Session

miloszm:
Moved deploy to session

Addresses issue #111

### DIFF
--- a/piecrust/src/module.rs
+++ b/piecrust/src/module.rs
@@ -7,6 +7,7 @@
 use crate::error::Error;
 use crate::instance::Store;
 
+#[derive(Clone)]
 pub struct WrappedModule {
     serialized: Vec<u8>,
 }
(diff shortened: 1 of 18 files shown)


piecrust/src/imports.rs
```diff
@@ -217,6 +217,7 @@ fn emit(mut fenv: FunctionEnvMut<Env>, arg_len: u32) {
     env.emit(arg_len)
 }
 
+#[allow(clippy::uninlined_format_args)]
```
krl, line 220: This does nothing on my machine, what is it for?

piecrust/src/module.rs
```diff
@@ -7,6 +7,7 @@
 use crate::error::Error;
 use crate::instance::Store;
 
+#[derive(Clone)]
```
krl, line 10: Use `Arc` instead of cloning the bytes.

krl (changes requested):

piecrust/src/imports.rs
miloszm: otherwise Clippy fails in the CI build

krl: Please provide the error or link to reproduction.

krl (approved):

Issue #113, opened 2022-10-27
Make `Session` borrow a mutable VM instead of cloning.


krl:
#### Summary
At the moment, the VM is cloned into the session, this makes the API a bit confusing, since it's unclear what commiting the session actually changes.

#### Possible solution design or implementation
Make `Session` take a `&mut VM` instead.

PR #115, opened 2022-10-27
Session and memory handler borrow VM immutably

miloszm:
Session and memory handler borrow VM immutably rather than clone the VM,
this PR also introduces an internal mutability for commits db in VM.

Addresses issue #113

### DIFF
--- a/piecrust/src/vm.rs
+++ b/piecrust/src/vm.rs
@@ -5,6 +5,7 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use std::borrow::Cow;
+use std::cell::RefCell;
 use std::collections::BTreeMap;
 use std::fmt::{self, Debug, Formatter};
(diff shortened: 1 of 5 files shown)


piecrust/src/vm.rs
```diff
@@ -29,7 +30,7 @@ const SESSION_COMMITS_FILENAME: &str = "commits";
 struct VMInner {
     host_queries: HostQueries,
     base_memory_path: PathBuf,
-    session_commits: SessionCommits,
+    session_commits: RefCell<SessionCommits>,
```
krl, line 33: Instead of using VMInner and a RefCell, just make the borrow &mut in  Session, as requested in the issue.

krl (changes requested):

OsDusk (approved):

krl (changes requested):
Remove VMInner and move members to VM directly

Issue #116, opened 2022-11-25
Add CheckBytes derivation to ModuleError


miloszm:
#### Summary
ModuleError enum has Archive derivation but lacks CheckBytes derivation, which is causing serialization to fail in Rusk.

#### Possible solution design or implementation
CheckBytes derivation needs to be added.

#### Additional context
All other Archive-derived structs and enums in Piecrust have CheckBytes derivation, so that is simply an omission which needs to be remedied.

PR #117, opened 2022-11-25
Added CheckBytes derivation to ModuleError

miloszm:
Added CheckBytes derivation to ModuleError - this is needed to fix Rusk serialization problem, at the same time it fixes the only struct/enum in Piecrust which has Archive derivation yet is lacking CheckBytes derivation.
Fixes issue #116


OsDusk (approved):

ureeves (approved):
LGTM

PR #121, opened 2022-12-16
Enforce a single instance per module

### LINKED ISSUE #122: `transact_raw` sometimes corrupts the stack on recursive calls
**Describe the bug**
Transact raw sometimes leads to stack corruption when a module calls itself recursively. This is visible in the `raw-calls-layout` branch in this repository. The `self_snapshot` test will produce an out of bounds access to the heap, which happens due to the pointer to the raw transaction buffer being corrupted.

**To Reproduce**
Take a look at the `raw-calls-layout` branch. An MRE exists there.

**Expected behaviour**
Quite simply this should not happen. A recursive call should not touch the caller's stack.

**Additional context**
This was found in a larger effort to migrate the genesis contracts to use this library. `transact_raw` is used there https://github.com/dusk-network/rusk/issues/757.

ureeves:
Ensures a module is only ever instantiated once per session, by using a map in the call stack. If an instance already exists in an instance map, then an element is appended without creating a new instance. If an instance does not exist in the map, a new one is created and then a new element appended to the stack.

Resolves #122

### DIFF
--- a/piecrust/src/session/call_stack.rs
+++ b/piecrust/src/session/call_stack.rs
@@ -1,0 +1,148 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use crate::instance::WrappedInstance;
+use std::collections::btree_map::Entry;
+
+use std::collections::BTreeMap;
+
+use piecrust_uplink::ModuleId;
+
+#[derive(Debug, Default)]
+pub struct CallStack {
+    // map of all instances together with a count in the stack.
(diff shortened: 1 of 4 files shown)


OsDusk (approved):

piecrust/src/session/call_stack.rs
```diff
@@ -0,0 +1,138 @@
...
+use std::collections::btree_map::Entry;
+
+use std::collections::BTreeMap;
+
+use piecrust_uplink::ModuleId;
+
+#[derive(Debug, Default)]
+pub struct CallStack {
+    // map of all instances together with a count in the stack.
+    instance_map: BTreeMap<ModuleId, (*mut WrappedInstance, u64)>,
```
ureeves, line 17: `CallStack` may drop without having released all the memory it contains. We should implement `Drop`

ureeves (commented):
One small problem before merging

PR #125, opened 2023-01-11
Cold Reboot Test and Commit Id Persistence

### LINKED ISSUE #124: Implement CommitId persistence and Cold Restart test in Piecrust
#### Summary
Cold restart functionality and test were present in Rusk-VM and we should provide equivalent functionality in Piecrust.

We need to add `from` and `persist` methods for CommitId, both accepting `AsRef<Path>` as a parameter. Alternative solution would be to implement similar functionality in Rusk, yet it would be less elegant and universal, as this functionality really belongs in Piecrust and its cold restart functionality, and may be used by other modules using Piecrust in the future.  Piecrust is currently lacking the cold restart test, which should be implemented in a way equivalent to the corresponding cold restart test in Rusk-VM.

#### Additional context
Conceptually, CommitId in Piecrust is an equivalent of NetworkStateId as defined in Rusk-VM and used in Rusk. When porting Rusk to Piecrust, we need CommitId to support persistence, in order for it to be used as conveniently as NetworkStateId in Rusk. Implementation of this issue should also include a test covering the VM cold restart, as in Rusk-VM's test_runner module.

miloszm:
Implements Cold Reboot Test and Commit Id Persistence

Fixes #124

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,4 @@
 members = [
     "piecrust-uplink",
     "piecrust",
-    "cold_reboot_test",
 ]
-
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,5 +2,6 @@
 members = [
     "piecrust-uplink",
     "piecrust",
+    "cold_reboot_test",
 ]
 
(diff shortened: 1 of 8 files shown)


ureeves (approved):
LGTM

Issue #132, opened 2023-01-12
Improve Piecrust Error Infrastructure to support conversions


miloszm:
#### Summary
Piecrust error needs to implement trait `std::error::Error` in order to be easily integrated with client's code.

#### Possible solution design or implementation
This can be done by using automatic derivation provided by `thiserror` crate.

#### Additional context
This feature is important for the (among other) Rusk-Recovery implementation. Note that VM1 a.k.a. rusk-vm supported VMError which did implement the desired trait so we have a regression in this regard here for Piecrust.

miloszm:
Fixed by PR 134

PR #134, opened 2023-01-13
Error Infrastructure Enhancement

miloszm:
Error Infrastructure Enhancement

Implements #132

### DIFF
--- a/piecrust/Cargo.toml
+++ b/piecrust/Cargo.toml
@@ -19,7 +19,7 @@ colored = "2"
 libc = { version = "^0.2", default-features = false }
 region = "3"
 tempfile = "3"
-thiserror = "1.0"
+thiserror = "1"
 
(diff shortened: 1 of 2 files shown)


piecrust/Cargo.toml
```diff
@@ -19,6 +19,7 @@ colored = "2"
 libc = { version = "^0.2", default-features = false }
 region = "3"
 tempfile = "3"
+thiserror = "1.0"
```
ureeves, line 22: Since the others above only specify major version, perhaps we should do the same with this one.

ureeves:
```suggestion
thiserror = "1"
```

ureeves (approved):
One nit, but otherwise LGTM

herr-seppia (approved):
LGTM

PR #137, opened 2023-01-13
Make `transact` and `query` take arguments by reference

### LINKED ISSUE #135: Call arguments should be taken by reference
#### Summary
Currently the argument for calls - be it transactions or queries - are taken by value. This means a consumer of the VM has to, if they want to reuse the argument after executing a call, clone the argument instance. This is undesirable since these object can be pretty large - up to 64kB. The good news is we can take it by reference since the argument is only serialized into the argument buffer.

The `Session::transact` and `Session::query` should take a reference to an `Arg` instead of consuming it. This is possible since the argument is only ever serialized into a buffer, hence consumption is not required.

Ideally, the functions available to module writers to perform inter-contract calls should also take the arguments by reference in `piecrust-uplink`.

https://github.com/dusk-network/piecrust/blob/e55c691e3c9f3203f504305d221a075a4dffa822/piecrust/src/instance.rs#L258-L273

#### Additional context

In the process of [migrating `rusk`](https://github.com/dusk-network/rusk/issues/759) to use this library, it was noticed that the cost of these functions consuming the argument is quite high, since the arguments used are not always `Copy`.

ureeves:
Resolves #135

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ let mut vm = VM::ephemeral()?;
 let module_id = vm.deploy(bytecode)?;
 
 let mut session = vm.session();
-let result = session.transact::<i16, i32>(module_id, "function_name", 0x11)?;
+let result = session.transact::<i16, i32>(module_id, "function_name", &0x11)?;
 
 // use result
 ```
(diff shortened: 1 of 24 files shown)


miloszm (approved):
LGTM

PR #141, opened 2023-01-17
VM state Merkle root calculation

### LINKED ISSUE #55: Construct a tree from module memories
One of the most important features the VM should provide is the so called "state root". The state root is the root of a [Merkle tree](https://en.wikipedia.org/wiki/Merkle_tree), where the leaves of the tree are the hashes of the bytecode of the module and its linear memory.
This root changes across snapshots, and would - optionally - make a great `SnapshotId`.

To ensure each participant in the blockchain can efficiently verify the state of their VM is the same as all other participants, the state root is included in a block. It is, therefore, a requirement.

An alternative way of providing such a state root would be to hash all modules sequentially. This is, however, computationally prohibitive for a large set of modules. I would also make make efficient inclusion proofs impossible.

**Additional context**
If `ModuleId` is used to determine the position in the tree, this tree will likely be quite sparse. The implementer will have to decide what the most sensible way to populate the tree for both computational and spatial efficiency.

miloszm:
VM state Merkle root calculation.
Extends persistence scheme, as described in on-disk-datastore.md.
Introduces generic Persistable trait.

Resolves #55


piecrust/src/vm.rs
```diff
@@ -15,17 +15,21 @@ use piecrust_uplink::ModuleId;
...
 
 const SESSION_COMMITS_FILENAME: &str = "commits";
+const LAST_COMMIT_POSTFIX: &str = "_last";
+const LAST_COMMIT_ID_POSTFIX: &str = "_last_id";
 
 pub struct VM {
     host_queries: HostQueries,
     base_memory_path: PathBuf,
     session_commits: SessionCommits,
+    root: Option<[u8; 32]>,
```
ureeves, line 32: This doesn't make sense here IMHO.

ureeves (changes requested):
I don't think that the VM struct should contain information about "latest state root". A state root IMO only makes sense in the context of a session.

ureeves (approved):
It was decided that we will keep this concept in the VM for now. Approved.

PR #147, opened 2023-01-27
Persist module bytecode as part of the state

### LINKED ISSUE #144: Automatic Module Redeployment after Session Restore
#### Summary
Information about which modules were deployed by the session is currently lost after the session is committed and destructed. When session is restored from a commit, user needs to manually deploy necessary modules. Automatic mechanism for redeployment is needed so that after restore the user of a session operate as if the session was not stored and restored.

Modules' bytecode need to be stored along with the commit data. The design and details of the solution needs to be elaborated and is part of this issue's scope. 

#### Additional context
Solution may involve introduction and use of a database in Piecrust. Modules' bytecode could be kept within the database together with the other commit data.

ureeves:
Resolves #144

### DIFF
--- a/piecrust/src/vm.rs
+++ b/piecrust/src/vm.rs
@@ -8,13 +8,15 @@ use std::borrow::Cow;
 use std::collections::{BTreeMap, HashSet};
 use std::fmt::{self, Debug, Formatter};
 use std::path::{Path, PathBuf};
+use std::{fs, io};
 
 use tempfile::tempdir;
 
-use piecrust_uplink::ModuleId;
+use piecrust_uplink::{ModuleId, MODULE_ID_BYTES};
 
 use crate::commit::{CommitId, ModuleCommitId, SessionCommit, SessionCommits};
 use crate::memory_path::MemoryPath;
+use crate::module::WrappedModule;
 use crate::persistable::Persistable;
 use crate::session::Session;
 use crate::types::MemoryState;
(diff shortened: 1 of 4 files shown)


piecrust/src/vm.rs
```diff
@@ -41,11 +114,14 @@ impl VM {
         let session_commits = SessionCommits::from(
             base_memory_path.join(SESSION_COMMITS_FILENAME),
         )?;
+        let modules = read_modules(&base_memory_path)?;
```
miloszm, line 117: I am a bit discomforted that VM's `new()` is doing actual directory scanning. Could we make an explicit step for scanning so that we have more control over it, IMO new() should be a lightweight function without performing IO.

piecrust/src/vm.rs
```diff
@@ -24,12 +26,83 @@ use crate::Error::{self, PersistenceError, RestoreError};
...
+    if module_id_bytes.len() != MODULE_ID_BYTES {
+        return Err(PersistenceError(io::Error::new(
+            io::ErrorKind::InvalidData,
+            format!(
+                "Expected file name of length {MODULE_ID_BYTES}, found {}",
+                module_id_bytes.len()
+            ),
+        )));
+    }
+
```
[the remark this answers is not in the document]
ureeves: Moved to `util` module

piecrust/src/vm.rs
ureeves: It seems to me that it doesn't make sense to instantiate a `VM` without knowledge of the modules it has deployed. Are you suggesting the instantiation process should look something like this: ?

```rust
let vm = VM::new(path)?;
vm.load()?;
```

This seems cumbersome and I can't see a usecase for just using `VM::new` then.

miloszm: I'd name it sth different than `new` then, otherwise I have no objections

miloszm (approved):
LGTM

Issue #152, opened 2023-02-12
Provide coverage for non-unique diffed module commit ids


miloszm:
#### Summary
Rusk's wallet_grpc test provides a scenario in which module commit ids are appearing repeatedly in a commit history. This requires diffed commits to be stored independently in the presence of multiple repeating ids and not overwrite each other.

#### Possible solution design or implementation
The implementation requires introducing a diff commit file naming scheme that is not mapped 1-1 with commit ids.

#### Additional context
The functionality needs to be performant in the presence of a large number of commits (say 200-1000). In addition, it will have impact in potential commit cut-off or squash schemes.

ureeves:
The conflict referenced is impossible since the introduction of the `ModuleStore`.

PR #153, opened 2023-02-13
Diffing Enhancement

miloszm:
This PR adds the following functionality to the diffing implementation, points 1 and 2 are needed to pass the wallet_grpc test in Rusk:
1) support for last commits of modules which are not present in current session
2) unique diff files
3) deleting original image of file when diff is done
4) improved root calculation and caching
Implements issue #152


ureeves (approved):
LGTM

PR #154, opened 2023-02-15
Support session concurrency with storage model

### LINKED ISSUE #145: Structure modules and memories to allow `Session` concurrency
#### Summary
To allow for multiple sessions to be executed in parallel, the way in which the modules are structured on disk needs to be changed to support it.

It should be possible to base sessions on top of different commits and have them execute concurrently, without interfering with each other.

This also means removing the concept of "current_commit" from the crate, since it no longer makes sense in the context of concurrent sessions.

Together with @miloszm, we came up with one possible structuring of the modules to enable concurrent sessions described in [this document](https://github.com/dusk-network/piecrust/files/10491991/piecrust.pdf). 

It is possible to flatten the structure a little bit, by removing the nested folders used for commits, memories, and bytecode and instead concatenating names in a structured way.

In the document, one thing that is introduced is commit ordering using a "base" file pointing to the base commit. This is also possible using a database file. This might be more efficient since it is likely that during certain operations - particularly collapsing/deleting commits - will require not only a "backwards" pointer, but also a "forward" pointer denoting children commits.

Implementing this will likely give rise to a different API for `Session` and `VM`, particularly on starting a session and committing it. APIs should also be exposed for deleting particular commit paths. One example is:

```rust
impl Session {
    pub fn commit(self) -> Result<CommitId> {
        // ...
    }
}

impl VM {
    pub fn session(&self, base_commit: Option<CommitId>) -> Result<Session> {
        // ...
    }

    pub fn collapse(&self, commit: CommitId) -> Result<()> {
        // ...
    }
}
```

Since a commit shouldn't be deleted or otherwise modified while a `Session` is basing itself on it, commit deletion/collapse and session spawning need to be synchronized.

#### Additional context
This is in the context of [`rusk` migration to piecrust](https://github.com/dusk-network/rusk/issues/760), and it is necessary to allow `rusk` full control of which commits are kept and which ones are discarded.

ureeves:
Resolves #145


piecrust/src/store.rs
```diff
@@ -0,0 +1,903 @@
...
+///
+/// A session tracks modifications to the modules' memories by keeping
+/// references to the set of instantiated modules.
+///
+/// The modifications are kept in memory and are only persisted to disk on a
+/// call to [`commit`].
+///
+/// [`commit`]: ModuleSession::commit
+pub struct ModuleSession {
+    modules: BTreeMap<ModuleId, (Bytecode, Memory)>,
```
miloszm, line 719: I'd consider moving ModuleSession to a separate file, as the class is important in its own right and also the file store.rs is long already.

piecrust/src/store.rs
```diff
@@ -0,0 +1,903 @@
...
+            diffs.insert(*module);
+        }
+    }
+
+    Ok(Commit { modules, diffs })
+}
+
+fn merkle_from_path<P: AsRef<Path>>(
+    path: P,
+) -> io::Result<BTreeMap<ModuleId, Root>> {
```
miloszm, line 262: I'd consider changing the name of this function as it is misleading, it returns a map of module roots rather than a single Merkle root

miloszm (approved):

PR #161, opened 2023-02-27
Implement call atomicity

### LINKED ISSUE #155: Crossover doesn't behave as expected
**Describe the bug**
If a contract fails - or panics - in its execution, the state that it has already modified will stay modified.

**To Reproduce**
Run the  `crossover-test-fix` branch to see the failing test.

**Expected behavior**
The changes performed by a contract in the context of it being called should rollback if the contract panics or otherwise signal failure.

**Additional context**
The crossover test was intended to mimic the behavior of the transfer and stake contracts while staking. It is important that a panic in the stake contract doesn't touch the state of the transfer contract. The same applies for technically all contracts.

### LINKED ISSUE #156: Events should not be emitted by panicked contracts
**Describe the bug**
Currently, even if a contract panics, the events it emitted will be kept in the events list. This should not be the case.

**To Reproduce**
Build a contract that emits an event and then panics.

**Expected behaviour**
Events should be removed from the list during a contract panic, or even not be added unless the contract succeeds in its execution.

**Additional context**
[#155](https://github.com/dusk-network/piecrust/issues/155)

ureeves:
Each call, be it a transaction or a query, is now atomic - meaning it
will not change the state if it fails. This is also true for every
inter-contract call performed during said call.
    
This is achieved by a re-execution strategy. Since we have no way of
partially rolling back changes in memory, each failing call - and
inter-contract for that matter - will trigger a re-execution of all
historic calls in a session until said failure occurs. The failure is
then skipped, communicated to the caller, and execution proceeds until
the next error - if so the procedure is gone through again - or until it
succeeds.
    
Resolves #155
Resolves #156

### DIFF
--- a/modules/crossover/src/lib.rs
+++ b/modules/crossover/src/lib.rs
@@ -10,7 +10,7 @@
 #![feature(core_intrinsics, lang_items, arbitrary_self_types)]
 
 use piecrust_uplink as uplink;
-use uplink::{ModuleId, RawTransaction, State};
+use uplink::{ModuleId, State};
 
 /// Struct that describes the state of the crossover module
 pub struct Crossover {
@@ -21,10 +21,59 @@ pub struct Crossover {
 #[no_mangle]
 static SELF_ID: ModuleId = ModuleId::uninitialized();
 
+const INITIAL_VALUE: i32 = 0;
+
 /// State of the crossover module
-static mut STATE: State<Crossover> = State::new(Crossover { crossover: 7 });
+static mut STATE: State<Crossover> = State::new(Crossover {
+    crossover: INITIAL_VALUE,
+});
 
 impl Crossover {
+    // Calls another contract - which is assumed to be another crossover
+    // contract - with the "set_and_panic" call.
+    //
+    // The other contract will first set their state, call this contract to set
+    // the crossover, and then panic.
+    //
+    // We then proceed to query the contract, and return true if both contract's
+    // states were unchanged. Before returning, the contract's state is set to
+    // the new `value`.
+    pub fn call_panicking_and_set(
+        self: &mut State<Self>,
+        module: ModuleId,
+        value: i32,
+    ) -> bool {
+        uplink::debug!("Calling panicking module {module:?}");
+        self.transact::<_, ()>(module, "set_call_and_panic", &value)
+            .unwrap_err();
+
+        let self_is_initial = self.crossover == INITIAL_VALUE;
+
+        uplink::debug!("Querying module {module:?} for their state");
+        let other_crossover =
+            uplink::query::<_, i32>(module, "crossover", &value).unwrap();
+        let other_is_initial = other_crossover == INITIAL_VALUE;
+
+        self.set_crossover(value);
+
+        self_is_initial && other_is_initial
+    }
+
+    // Sets the crossover to a new `value`, calls the calling contract - which
+    // is assumed to be another crossover contract - with "set_crossover" and
+    // panics afterwards.
+    pub fn set_call_and_panic(self: &mut State<Self>, value: i32) {
+        self.set_crossover(value);
+
+        let caller = uplink::caller();
+        uplink::debug!("calling back {caller:?}");
+        self.transact::<_, ()>(caller, "set_crossover", &value)
+            .unwrap();
+
+        uplink::debug!("panicking after setting the crossover");
+        panic!("OH NOES");
+    }
+
     /// Return crossover value
     pub fn crossover(&self) -> i32 {
         self.crossover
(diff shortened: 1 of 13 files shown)


piecrust-uplink/src/state.rs
```diff
@@ -181,6 +214,10 @@ pub fn query_raw(
         ext::q(&mod_id.as_bytes()[0], &name[0], name.len() as u32, arg_len)
     };
 
+    if ret_len < 0 {
+        return Err(ModuleError::from_code(ret_len));
```
ureeves, line 218: Figure out if this is being correctly called

ureeves:
To be clear, this is an *extremely* expensive solution. I would prefer if we made some sort of OS backed snapshot mechanism, but this would likely entail 3 separate kernel modules for 3 separate operating systems. Maintaining these would be quite the task.

fed-franz:
> To be clear, this is an _extremely_ expensive solution. I would prefer if we made some sort of OS backed snapshot mechanism, but this would likely entail 3 separate kernel modules for 3 separate operating systems. Maintaining these would be quite the task.

Do we have any benchmark on this? I assume the worst case scenario entails re-executing the whole transaction but I can't get the feeling of how how _extreme_ the overhead is in terms of CPU time.

I think we should make a backlog Issue as a reminder to think about this and (hopefully) come up with alternative solutions (I think kernel modules would be a real pain to maintain).

ureeves:
> Do we have any benchmark on this? I assume the worst case scenario entails re-executing the whole transaction but I can't get the feeling of how how _extreme_ the overhead is in terms of CPU time.

It's worse than that. The worst case scenario entails re-executing all *previous calls* as well, since we can't be sure that they haven't encroached on each other's changes to the state.

We don't have a benchmark, and I hesitate in making one since we have no different approach to compare it to. Yet.

modules/crossover/src/lib.rs
```diff
@@ -21,66 +21,71 @@ pub struct Crossover {
...
-            _ => (),
-        }
+    /// Update crossover and return old value
+    pub fn set_crossover(&mut self, to: i32) -> i32 {
+        let old_val = self.crossover;
+        uplink::debug!("setting crossover from {old_val} to {to}");
+        self.crossover = to;
+        old_val
     }
 }
```
miloszm, line 91: As discussed, I'd use in this test contract different numbers so that it is easier and more convincing to follow state changes.

piecrust/tests/crossover.rs
```diff
@@ -4,44 +4,48 @@
...
+
+    assert!(
+        state_is_ok,
+        "The state should be unchanged in the panicking call's ray"
+    );
+    assert_eq!(
+        session.query::<_, i32>(CROSSOVER_ONE, "crossover", &())?,
+        CROSSOVER_TO_SET,
+        "The state should be properly set even after a call panics"
+    );
```
miloszm, line 49: See my remark in the crossover/src/lib.rs

miloszm (approved):
Nit remark regarding the crossover test, otherwise LGTM.

fed-franz (approved):
LGTM
Left a few suggestions

modules/crossover/src/lib.rs
ureeves: I made it a little more clear by allowing the caller to pass all different values to (possibly) set.

modules/crossover/src/lib.rs
```diff
@@ -21,66 +21,71 @@ pub struct Crossover {
...
+    // We then proceed to query the contract, and return true if both contract's
+    // states were unchanged. Before returning, the contract's state is set to
+    // the new `value`.
+    pub fn call_panicking_and_set(
+        self: &mut State<Self>,
+        module: ModuleId,
+        value: i32,
+    ) -> bool {
+        uplink::debug!("Calling panicking module {module:?}");
+        self.transact::<_, ()>(module, "set_call_and_panic", &value)
```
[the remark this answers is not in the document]
ureeves: Added a comment in the test that the number should never be set to `INITIAL_VALUE`.

modules/crossover/src/lib.rs
```diff
@@ -21,66 +21,71 @@ pub struct Crossover {
...
-    }
+    // Calls another contract - which is assumed to be another crossover
+    // contract - with the "set_and_panic" call.
+    //
+    // The other contract will first set their state, call this contract to set
+    // the crossover, and then panic.
+    //
+    // We then proceed to query the contract, and return true if both contract's
+    // states were unchanged. Before returning, the contract's state is set to
+    // the new `value`.
```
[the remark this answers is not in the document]
ureeves: Nice, took it over with some small modifications to better reflect some changes in the contract.

piecrust/src/session.rs
```diff
@@ -284,15 +317,245 @@ impl Session {
...
+            Entry::Occupied(mut entry) => {
+                let map = entry.get_mut();
+                map.insert(self.icc_count, err);
+            }
+        }
+    }
+
+    /// Execute the call and re-execute until the call errors with only itself
+    /// in the call stack.
+    fn re_execute_until_ok(&mut self, call: Call) -> Result<Vec<u8>, Error> {
```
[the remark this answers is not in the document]
ureeves: Renamed

piecrust/src/imports.rs
```diff
@@ -41,6 +42,8 @@ fn caller(env: FunctionEnvMut<Env>) {
         .nth_from_top(1)
         .map_or(ModuleId::uninitialized(), |elem| elem.module_id);
 
+    println!("Caller Module ID: {:?}", mod_id);
```
miloszm, line 45: Nit: this println needs to be removed.

PR #164, opened 2023-03-02
Added module caching

### LINKED ISSUE #162: push_callstack not storing module instances
**Describe the bug**
looks like every time push_callstack is called, a new module instance is created as existing instance is not found in instance_map

**To Reproduce**
run `cargo t counter_read_write_simple` and observe that module instance with the same module id is created 3 times

**Expected behaviour**
Module instance for a given module id should only be crated once as instance_map should cache instances

**Logs/Screenshot**
instance result from s.instance(e34e2104783737e9e747230e388638e53457fc82769b5c925732f10e321fd6de is some=false)
new instance called for module_id=e34e2104783737e9e747230e388638e53457fc82769b5c925732f10e321fd6de
creating wasmer module - bytecode len=443
instance result from s.instance(e34e2104783737e9e747230e388638e53457fc82769b5c925732f10e321fd6de is some=false)
new instance called for module_id=e34e2104783737e9e747230e388638e53457fc82769b5c925732f10e321fd6de
creating wasmer module - bytecode len=443
instance result from s.instance(e34e2104783737e9e747230e388638e53457fc82769b5c925732f10e321fd6de is some=false)
new instance called for module_id=e34e2104783737e9e747230e388638e53457fc82769b5c925732f10e321fd6de
creating wasmer module - bytecode len=443

**Platform**
MacOS

**Context**
On every session.query and/or session.transact new instance is created, this needs to be fixed.

miloszm:
Added module caching to avoid compilation on every query/transaction

Fixes #162

### DIFF
--- a/piecrust/src/error.rs
+++ b/piecrust/src/error.rs
@@ -41,6 +41,8 @@ pub enum Error {
     #[error(transparent)]
     CompositeSerializerError(Arc<Compo>),
     #[error(transparent)]
+    ModuleCacheError(Arc<std::io::Error>),
+    #[error(transparent)]
     PersistenceError(Arc<std::io::Error>),
     #[error("Commit error: {0}")]
     CommitError(Cow<'static, str>),
@@ -107,12 +109,6 @@ impl From<wasmer_vm::Trap> for Error {
 impl<A, B> From<rkyv::validation::CheckArchiveError<A, B>> for Error {
     fn from(_e: rkyv::validation::CheckArchiveError<A, B>) -> Self {
         Error::ValidationError
-    }
-}
-
-impl From<std::io::Error> for Error {
-    fn from(e: std::io::Error) -> Self {
-        Error::PersistenceError(Arc::from(e))
     }
 }
 
--- a/piecrust/src/error.rs
+++ b/piecrust/src/error.rs
@@ -110,6 +110,12 @@ impl<A, B> From<rkyv::validation::CheckArchiveError<A, B>> for Error {
     }
 }
 
+impl From<std::io::Error> for Error {
+    fn from(e: std::io::Error) -> Self {
+        Error::PersistenceError(Arc::from(e))
+    }
+}
+
 const OTHER_STATUS_CODE: i32 = i32::MIN;
 
 impl From<Error> for ModuleError {
--- a/piecrust/src/session/call_stack.rs
+++ b/piecrust/src/session/call_stack.rs
@@ -4,16 +4,10 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-use std::collections::btree_map::{BTreeMap, Entry};
-
 use piecrust_uplink::ModuleId;
-
-use crate::instance::WrappedInstance;
 
 #[derive(Debug, Default)]
 pub struct CallStack {
-    // map of all instances together with a count in the stack.
-    instance_map: BTreeMap<ModuleId, (*mut WrappedInstance, u64)>,
     stack: Vec<StackElement>,
 }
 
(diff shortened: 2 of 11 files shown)


ureeves (changes requested):
I don't think we should memoize calls to the compilation of each module. Instead, we should compile the bytecode on deployment, and store the result on disk such that it can be reloaded at any point without having to recompile it.

See the documentation of a [`wasmer::Module`](https://docs.rs/wasmer/latest/wasmer/struct.Module.html#method.serialize) for details on how this would work.

To be clear, I don't think this is wrong, just that there is a much cleaner way of achieving this - even without having to re-compile every cold reboot. Also, no new crates will be necessary.

piecrust/src/session/call_stack.rs
```diff
@@ -65,36 +64,14 @@ impl CallStack {
...
-
-            if *count == 0 {
-                // SAFETY: This is the last instance of the module in the
-                // stack, therefore we should recoup the memory and drop
-                //
-                // Any pointers to it will be left dangling
-                unsafe {
-                    let _ = Box::from_raw(*instance);
-                    entry.remove();
-                };
```
ureeves, line 95: If an instance is never removed from the map and dropped, then the linear memory `mmap` it is holding is never reloaded from disk - this is why the crossover test is failing

piecrust/src/error.rs
```diff
@@ -109,6 +109,12 @@ impl<A, B> From<rkyv::validation::CheckArchiveError<A, B>> for Error {
     }
 }
 
+impl From<std::io::Error> for Error {
+    fn from(e: std::io::Error) -> Self {
+        Error::PersistenceError(Arc::from(e))
+    }
+}
+
```
ureeves, line 117: Is an `io::Error` always a persistence error?

ureeves (commented):
Found the issue. Explanation in the comments. 

Also, why not just create a new file in the bytecode directory in each commit, representing the compiled bytecode?

Suggestion: Given a module named `module_1`:

```text
commit_dir/
    bytecode/
        module_1
        module_1.a # this being the compiled version
```

This would of course also need to be linked to in commits also using this

miloszm:
Regarding the above proposal of placement for compiled files, this is fine but it will lead to proliferation of links, as every compiled file will need to be multiply linked. Since this information is read only, wouldn't it be simpler to have a central location for it?

ureeves:
> Regarding the above proposal of placement for compiled files, this is fine but it will lead to proliferation of links, as every compiled file will need to be multiply linked. Since this information is read only, wouldn't it be simpler to have a central location for it? Yet I understand the desire of every commit to be fully standalone and independent.

This is true, but it's exactly the same issue as already exists, just that the number of files is increased by one. They would be of the exact same nature and behavior as the bytecode itself.

miloszm:
I have no problem with the above remark, issue is of different nature. If we want to use `serialize_to_file` method of Wasmer module, at the time this method is called, we don't have a commit yet, so basically at this time we don't have a path for the file. This leads to solutions based on `serialize` only, then we need to keep this buffer somewhere and pass it to commit routine as only at commit we know that path. I am not sure this is worth the complication but we can discuss.

miloszm:
One more explanatory remark - problem is that before we commit we do not know the path, we only know the path of previous commit or none, if this is a genesis session. With bytecode and module memory this is resolved so that we keep them in memory until the time of commit. With compiled object code this can also be done, but it is not ideal as it complicates the overall design and makes it impossible to use `serialize_to_file` method of Wasmer.

miloszm:
Changes making compilation-at-deploy-time and object code being stored along with the byte code (as per Ed's comment above), i.e., within the commit-related persistence infrastructure (constituting an object code caching mechanism) are implemented and ready to be reviewed. Given module needs to be compiled only once. Each commit will contain a hard link to object code resulting from compilation, and object code can be loaded (instantiated) on demand if needed, but not more often than once per session.

piecrust/src/store/session.rs
```diff
@@ -165,24 +173,15 @@ impl ModuleSession {
...
-        let module_id = ModuleId::from_bytes(hash.into());
-        self.deploy_with_id(module_id, bytes)?;
-
-        Ok(module_id)
+    /// Checks if module is deployed
+    pub fn module_deployed(&mut self, module_id: ModuleId) -> bool {
+        if self.modules.contains_key(&module_id) {
+            true
+        } else if let Some((_base, base_commit)) = &self.base {
+            base_commit.modules.contains_key(&module_id)
```
ureeves, line 181:
```suggestion
        } else if let Some((_, base_commit)) = &self.base {
            base_commit.modules.contains_key(&module_id)
```

piecrust/src/store.rs
```diff
@@ -34,6 +36,7 @@ const BYTECODE_DIR: &str = "bytecode";
 const MEMORY_DIR: &str = "memory";
 const DIFF_EXTENSION: &str = "diff";
 const INDEX_FILE: &str = "index";
+const OBJECTCODE_POSTFIX: &str = ".a";
```
ureeves, line 39:
```suggestion
const OBJECTCODE_EXTENSION: &str = ".a";
```

piecrust/src/store/session.rs
```diff
@@ -129,15 +129,22 @@ impl ModuleSession {
                             let base_dir = self.root_dir.join(base_hex);
 
                             let module_hex = hex::encode(module);
+                            let mut objectcode_name = module_hex.clone();
+                            objectcode_name.push_str(OBJECTCODE_POSTFIX);
```
ureeves, line 133: Try to use `Path::with_extension`

Issue #93, opened 2022-10-11
Module constructors


ureeves:
#### Summary
There should be a way to provide modules some data on deployment, which is then used to initialize the state. Currently it is only possible to "bake in" the initial state in the WASM bytecode. This precludes the module writer from writing a module once, and then reusing the same bytecode with different initial states. This use-case should become possible.

#### Possible solution design or implementation
It could be possible to check the existence of a such a module constructor by interpreting the module exports. If an export exists with a predefined name and signature, this could then be called with a piece of data provided by the deployer through the argument buffer. An example name and signature could be:

```rust
#[no_mangle]
fn ctor(pos: i32) {
    // ...
}
```

The `pos` argument would be the length of the argument in the buffer. The contract would then deserialize (possibly using `rkyv`) the piece of data and proceed with initialization.

This **has** to be accompanied with a mechanism that disallows the constructor to be called more than once.

#### Additional context
Not only is this a useful mechanism for prospective smart contract writers, it's also extremely useful to construct multiple networks with different types of genesis data, given that different networks are defined by differently initialized genesis contracts.

miloszm:
name should be `new` or `instantate`, arguments  are optional

ureeves:
Closed by https://github.com/dusk-network/piecrust/pull/168

PR #168, opened 2023-03-08
Module Constructor

miloszm:
Implements module constructor with dedicated test contract and test,
also renames host_data to meta_data and makes meta_data members optional.

Implements issue #93


modules/constructor/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+#[no_mangle]
+unsafe fn increment(arg_len: u32) -> u32 {
+    uplink::wrap_transaction(arg_len, |_: ()| STATE.increment())
+}
+
+/// Expose `Constructor::init()` to the host
+#[no_mangle]
+unsafe fn init(arg_len: u32) -> u32 {
+    uplink::wrap_transaction(arg_len, |arg: u8| STATE.init(arg))
+}
```
ureeves, line 61: I'm unsure about the `init` name. Proposals would be:

- `new`
- `ctor`
- `init`

ureeves (changes requested):
I'm not sure we should return `None` on *any* piece of metadata. If metadata is unset and the contract requests it, I think it should panic - much like de-referencing a pointer to memory not owned by a program.

miloszm:
Optionality of metadata was requested by Matteo, I noted down his request in issue #93. As far as name concerned, I am open for discussion.

piecrust/src/lib.rs
```diff
@@ -118,7 +118,7 @@ mod types;
 mod vm;
 
 pub use error::Error;
-pub use session::Session;
+pub use session::{Session, CONTRACT_INIT_METHOD};
```
ureeves, line 121: What is the reasoning for exporting this?

piecrust/src/session.rs
```diff
@@ -116,6 +117,35 @@ impl Session {
...
+    /// the arguments provided.
+    ///
+    /// If one needs to specify the ID, [`deploy_with_id`] is available.
+    /// If one would like to *not* call the initialization method, [`deploy`] is
+    /// available.
+    ///
+    /// [`ModuleId`]: ModuleId
+    /// [`deploy_with_id`]: `Session::deploy_with_id`
+    /// [`deploy`]: `Session::deploy`
+    pub fn deploy_and_init<Arg>(
```
ureeves, line 136: It's probably better to modify the existing function `deploy` and `deploy_with_id` to take an `Option<Arg>`. Doing it this way implies the caller should somehow "know" that the given bytecode has a constructor,, which they are clearly not in a position to do.

piecrust/src/session.rs
```diff
@@ -148,6 +178,42 @@ impl Session {
...
+    /// the arguments provided.
+    ///
+    /// If one would like to *not* specify the `ModuleId`, [`deploy`] is
+    /// available.
+    /// If one would like to *not* call the initialization method,
+    /// [`deploy_with_id`] is available.
+    ///
+    /// [`deploy`]: `Session::deploy`
+    /// [`deploy_with_id`]: `Session::deploy_with_id`
+    pub fn deploy_with_id_and_init<Arg>(
```
ureeves, line 192: Ditto here

ureeves (changes requested):
Overall the implementation is good, but there's several issues I can see. I don't think we need to broaden the exposed API for this -  meaning we should avoid the proliferation of `deploy-*` calls, exporting constants (only used in tests).

There's also the problem of call history. With this implementation, an `init` call cannot be replayed, since we forget the data used in the call.

piecrust/tests/box.rs
```diff
@@ -12,7 +12,7 @@ pub fn box_set_get() -> Result<(), Error> {
 
     let mut session = vm.genesis_session();
 
-    let id = session.deploy(module_bytecode!("box"))?;
+    let id = session.deploy(module_bytecode!("box"), None::<()>)?;
```
ureeves, line 15: I didn't know one could do this on the enum's variant. Interesting!

ureeves (changes requested):
I think the serialization logic can be further simplified. Below a suggestion on how to do that, following the same pattern as what `query` and `transact` do.

ureeves (approved):
LGTM

Issue #158, opened 2023-02-24
Ownership of Contracts


miloszm:
#### Summary
We need to support the concept of contract ownership. 
Ownership of a contract may help deciding who is allowed to execute certain contract methods.

#### Possible solution design or implementation
Ownership information is passed via constructor arguments and maintained by contract.

#### Additional context
Contract ownership should be passed to a contract that is deployed (instantiated) by a given contract.

Rename host_data to meta_data, don't panic if it does not exist, return option.
Identifier needs to be accessible in the top level transaction and the transactions initiated by it.
Identifier is a hash of a module id and its byte-code. From the point of view of Piecrust, this so called instance identifier is an opaque object which will only be compared but nothing needs to be known about its internal structure.

Related to issue #93

miloszm:
Implementation for this issue is covered by the implementation for issue #167.

ureeves:
Close by https://github.com/dusk-network/piecrust/pull/180

PR #170, opened 2023-03-14
Made meta data access option-based, never panicking

miloszm:
Host data has been renamed o meta data, access to which
is now optional, never panicking.

Described in issue #158


ureeves (approved):
LGTM

Issue #167, opened 2023-03-08
Secure Persistent Metadata for Contracts


miloszm:
#### Summary
For contracts attributes such as contract's owner or contract's SELF_ID we need the ability of storing and accessing such attributes on a per module (contract) basis. A set of such attributes constitute contract's metadata. Distinguishing feature of contract's metadata is that it is managed by the VM and from the point of view of the contract it is immutable. Another way to look at it is that metadata constitutes a context for the contract.

#### Possible solution design or implementation
We store contract's metadata in a file in a similar manner to how object code is stored.

#### Additional context
This issue is related to issue #158 (ownership of contracts) which requires storing of contract's owner.

ureeves:
Closed by https://github.com/dusk-network/piecrust/pull/180

PR #180, opened 2023-03-27
Metadata for Contracts

miloszm:
Implementation of Secure Persistent Metadata for Contracts.

Also implements support for contract owner and contract self_id.
The self_id part does not require contracts to declare SELF_ID element any more,
so SELF_ID related functionality and declarations have been removed.
uplink::owner() and uplink::self_id() have been added to the uplink interface.
deploy API has beed updated to pass owner and self_id.

Implements #167
Related to #158

### DIFF
--- a/piecrust/CHANGELOG.md
+++ b/piecrust/CHANGELOG.md
@@ -13,14 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Removed SELF_ID export from contracts [#167]
-- Added uplink::owner and uplink::self_id [#158]
-- Added persistence fo module metadata [#167]
-- Changed deploy API to accept self_id and owner [#158] 
-- Added contract constructor mechanism [#93]
-- Added caching of module compilation outputs [#162]
-- Made modules compile at deploy time rather than on first query/transaction time [#162]
-- Removed errant print statements
+- Removed SELF_ID export from contracts. [#167]
+- Added uplink::owner and uplink::self_id. [#158]
+- Added persistence fo module metadata. [#167]
+- Added `DeployData` and `DeployDataBuilder`. [#158]
+- Changed deploy API to accept `Into<DeployData>`. [#158]
+- Added contract constructor mechanism. [#93]
+- Added caching of module compilation outputs. [#162]
+- Made modules compile at deploy time rather than on first query/transaction time. [#162]
+- Removed errant print statements.
 
 ## [0.1.0] - 2023-03-15
 


piecrust/tests/cold-reboot/src/main.rs
```diff
@@ -26,7 +27,7 @@ fn initialize_counter<P: AsRef<Path>>(
         "../../../../target/wasm32-unknown-unknown/release/counter.wasm"
     );
 
-    session.deploy_with_id(COUNTER_ID, counter_bytecode, None::<&()>)?;
+    session.deploy(counter_bytecode, ModuleData::new(Some(COUNTER_ID.to_bytes()), None::<()>, OWNER))?;
```
ureeves, line 30: I think `ModuleId` is `Copy`, right? Why is the `to_bytes()` needed?

ureeves (changes requested):
I think this is pretty good already, but there's some things I would like to discuss.

piecrust/src/session.rs
```diff
@@ -762,6 +786,7 @@ struct Deploy {
     module_id: ModuleId,
     bytecode: Vec<u8>,
     fdata: Option<Vec<u8>>,
+    owner: [u8; 32],
```
ureeves, line 789: Maybe we should include the whole `DeployData` struct here instead of `owner`, `fdata`, and `module_id` separately.

piecrust/src/store/metadata.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use crate::module::ModuleMetadata;
+use crate::store::mmap::Mmap;
+
+/// Module metadata pertaining to a given module but maintained by the host.
+#[derive(Debug, Clone)]
+pub struct Metadata {
+    mmap: Arc<Mmap>,
+    data: ModuleMetadata,
+}
```
ureeves, line 21: It doesn't seem that we need to keep the `mmap` field here. Once it is deserialized into memory we can discard the mmap, no?

ureeves: And if this is the case, this struct is unnecessary, since the `from_file` function could be on `ModuleMetadata` instead.

piecrust/src/store/module_session.rs
```diff
@@ -10,13 +10,55 @@ use std::path::{Path, PathBuf};
...
+    OBJECTCODE_EXTENSION,
 };
 
+#[derive(Debug, Clone)]
+pub struct ModuleData {
+    bytecode: Bytecode,
+    objectcode: Objectcode,
+    metadata: Metadata,
+    memory: Memory,
+}
```
ureeves, line 28: Since this is merely a piece of data that lives together, like the others before it, and it is also internal, I feel like making fields `pub` and removing accessors and `new` is fine. Maybe I'm wrong about this but it feels like we're trying to model this as an object when it's just plain data.

ureeves (changes requested):
Some possible changes to be discussed, nothing major.

piecrust/src/store/metadata.rs
miloszm: Problem is that we are inserting Metadata to ModuleData, so we need to have `from_file` which produces Metadata. With `mmap` elimination - I don't see how to do it without extra serializations. Issue is that Metadata is needed in serialized form as it is being stored along with ModuleData, so if we eliminate `mmap`, we end up serializing on demand which will be wasteful.

piecrust/src/session.rs
miloszm: Problem is that DeployData has a type argument `DeployData<Arg>` and this leads to a ripple effect leading to code that does not have the `Arg` available. This will require major restructuring and adding the `Arg` in lots of places. I am not sure it is worth it.

ureeves: You're right here. Not worth it.

piecrust/CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Removed SELF_ID export from contracts [#167]
+- Added uplink::owner and uplink::self_id [#158]
+- Added persistence fo module metadata [#167]
+- Changed deploy API to accept self_id and owner [#158]
```
ureeves, line 19: This line needs to change according to the new builder

ureeves (approved):
LGTM. Be sure to adjust the CHANGELOG though

piecrust/src/store/metadata.rs
ureeves: I think we can tackle this in #181

Issue #178, opened 2023-03-22
ModuleId should implement Display


miloszm:
#### Summary
We should not need to use hex::encode() in places where Display trait is required. Display should be implemented for ModuleId converting it into a hex encoded string in a similar manner to hex::encode.

PR #183, opened 2023-03-31
Implement Display for ModuleId

miloszm:
Implement Display for ModuleId

This is needed in Rusk.

Implements #178


HDauven (approved):
LGTM

Would it make sense to add a test that checks if the `ModuleId` returned by a session deploy is not empty and adheres to the specification of the `ModuleId`? (Length for example)

This could be done by adding an assert in one of the test contracts for example.

miloszm:
Hein - yes, in general, we miss more of such types of tests, checking the integrity of the system, I will add an issue for that.

Issue #181, opened 2023-03-29
Better structure and data encapsulation


ureeves:
#### Summary
In the context of implementing #180 (Issue #167), we have noticed that there are several concepts that can be grouped together in particular ways to better structure the project, reduce the maintenance burden, and aid in the development of new features.

In particular, there seem to be at least three different categories of data:

- Module data. This is data that is set at module deployment time, and never changes in the lifetime of the module. Examples are:
    - Constructor data (transient/not persisted)
    - Ownership data
    - Self ID

An API should be exposed that hints at, or even enforces, that these things cannot be changed for the lifetime of the module, possibly at deploy time.

- Session data. This is data that is set at the beginning of the session, and never changed for its duration. Examples include:
    - What is currently called `SessionMetadata` - contains universal settable map containing context items like 'height' 

In the same way as before, an API should be exposed that hints, or enforces, that these pieces of data are set at the beginning of a session, possibly during instantiation.

- Call data. This is data that is set at the beginning of a call to `transact` or `query`. This is data such as:
    - (Gas) limit
    - Sender (same entity as Ownership)

This should be set when a call is performed.

#### Possible solution design or implementation
A clear solution would be to give each of these structs, and allow them to be set only during the appropriate time:

```rust
struct ModuleData {
    pub self_id: Option<ModuleId>,
    pub constructor: Option<?>,
    pub owner: [u8; 32],
}

impl Session {
    fn deploy(bytecode: &[u8], data: ModuleData) {
        // ...
    }
}

struct SessionData {
    map: Map// some map from string to `Vec<u8>`
}

impl VM {
    fn session(root: [u8; 32], data: SessionData) -> Session {
        // ...
    }
}

struct CallData {
    pub limit: u64,
    pub sender: [u8; 32],
}

impl Session {
    fn transact<Arg, Ret>(module_id: ModuleId, method: &str, arg: Arg, data: CallData) -> Ret {
        // ...
    }
}
```

It is important to note that both the API and the structures outlined here are mere suggestions, and it is up to the implementer to best structure this.

miloszm:
It looks like module data has a proper API already, it can be set only via the `deploy` call and it is read-only after a module is deployed. It is different with session data - API allows for setting it any time so this part needs work - elimination of the `set_meta` call and extending the session creation API with the possibility of passing session data which would, after session creation, be only accessible as read-only data.

miloszm:
[...]

PR #185, opened 2023-04-04
Structure improvement and data encapsulation

miloszm:
This PR implements issue #181. Implementation consists of the following parts:

1) (ModuleData) Renamed DeployData as ModuleData. Internal ModuleData has been renamed to ModuleDataEntry. Renamed session's method metadata to module_metadata.

2) (SessionData) Eliminated the set_meta API, enriched he session creation API (VM's methods session and genesis_session) as well as deployment API with the possibility of passing session data. Said session data, after session creation, is made accessible only as read-only.

3) (CallData) Is implemented in a separate branch to allow proper evaluation and discussion, otherwise the changes are too voluminous for easy evaluation. Branch name is `structure-improv-call-data-181`. Unlike points 1 and 2, I am not sure if changes for point 3 should be merged.

### DIFF
--- a/piecrust/src/vm.rs
+++ b/piecrust/src/vm.rs
@@ -106,10 +106,10 @@ impl VM {
     ///
     /// [`Session`]: Session
     /// [`genesis_session`]: VM::genesis_session
-    pub fn session<D>(&self, data: D) -> Result<Session, Error>
-    where
-        D: Into<SessionData>,
-    {
+    pub fn session(
+        &self,
+        data: impl Into<SessionData>,
+    ) -> Result<Session, Error> {
         let data = data.into();
         let module_session = match data.base {
             Some(base) => self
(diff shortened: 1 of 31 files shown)


piecrust/src/vm.rs
```diff
@@ -125,9 +133,9 @@ impl VM {
     ///
     /// [`Session`]: Session
     /// [`session`]: VM::session
-    pub fn genesis_session(&self) -> Session {
+    pub fn genesis_session(&self, data: SessionData) -> Session {
```
ureeves, line 136: Could the commit `base` also be in `SessionData`? In this way, we could even get rid of the `genesis_session()` in favor of just `session()`. It also starts feeling like the same builder pattern as `ModuleData` should be applied here.

ureeves (changes requested):
Some comments and one thing to be discussed.

ureeves (approved):
LGTM. Really small nits

Issue #166, opened 2023-03-08
Store the merkle tree of contracts


ureeves:
#### Summary
Currently we use the `index` file to store the leaves of the tree, but we never store the full tree - meaning the leaves together with all the nodes. This would be useful for increasing the speed of root calculations, as well a easy inspection of the state integrity.

#### Possible solution design or implementation
It might be possible to use the `nstack` for this, but it might be useful to look into trees with arbitrary insertion capabilities, as opposed to append only.

#### Additional context
If using `nstack`, we would could change the order of the leaves to deployment order, rather than module ID order. This would be a breaking change.

ureeves:
This issue is being tackled by making a new merkle tree @
https://github.com/dusk-network/merkle

PR #187, opened 2023-04-11
Store Merkle tree of modules on disk

ureeves:
We use `dusk-merkle` to represent the merkle tree, and update it when a new commit is written to disk. The hash chosen for the leaves is `blake3`.

Resolves: #166

### DIFF
--- a/piecrust/src/store/tree.rs
+++ b/piecrust/src/store/tree.rs
@@ -1,0 +1,160 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use bytecheck::CheckBytes;
+use piecrust_uplink::ModuleId;
+use rkyv::{Archive, Deserialize, Serialize};
+
+// This means we have max `2^32` modules
+const HEIGHT: usize = 32;
+const ARITY: usize = 2;
+
+pub type Tree = dusk_merkle::Tree<Hash, HEIGHT, ARITY>;
+
+#[derive(
+    Debug,
+    Clone,
+    Copy,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    Archive,
+    Deserialize,
+    Serialize,
+    CheckBytes,
+)]
+#[archive(as = "Self")]
+pub struct Hash([u8; blake3::OUT_LEN]);
+
+impl Hash {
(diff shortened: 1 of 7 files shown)


piecrust/src/store/tree.rs
```diff
@@ -0,0 +1,151 @@
...
+    pub fn update(&mut self, input: &[u8]) -> &mut Self {
+        self.0.update(input);
+        self
+    }
+
+    pub fn finalize(&self) -> Hash {
+        let hash = self.0.finalize();
+        Hash(hash.into())
+    }
+}
```
moCello, line 91: What's the reason behind the creation of this `Hasher` type? Why not use `blake3::Hasher` directly?

moCello (approved):
LGTM

piecrust/src/store/tree.rs
ureeves: In the future it would be nice to be able to change hash function by just targeting a single file. I should probably make these functions `inline(always)` though.

piecrust/src/store/tree.rs
```diff
@@ -0,0 +1,154 @@
...
+
+/// Returns the position of a in the tree `module` given its ID.
+///
+/// # SAFETY:
+/// If there ever is a collision between IDs this is likely the culprit and
+/// should be changed.
```
fed-franz, line 100: Maybe add a short description of how we compute the position form ModuleId.
e.g.:

To compute the position in the Module Tree, the 32-byte ModuleId is divided in 8 4-byte slices, which are summed up together (modulo 2^32) to obtain the position value.

Also, the SAFETY message should probably be removed or updated, since we prevent collisions when inserting a module. We should mention this here.

e.g.: 
SAFETY: since we are mapping 32-byte values to 4-byte values, it is possible to have collisions between different `ModuleId`s. To prevent this, we check for collisions before inserting a new module (see `deploy`).

fed-franz (commented):
Fine overall, but I would add the following:
- a description of the Dusk Merkle tree (even better, justifying why we use our own implementation)
- a mention, in store.rs, of the fact that we store modules in a (dusk_merkle) tree
- a description of the mapping function (position_from_module), as per comment

piecrust/src/store/tree.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    #[inline(always)]
+    pub fn finalize(&self) -> Hash {
+        let hash = self.0.finalize();
+        Hash(hash.into())
+    }
+}
+
+/// Returns the position of a in the tree `module` given its ID. The position is
```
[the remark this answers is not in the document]
ureeves: What in the world happened there... Hahaha

Issue #197, opened 2023-04-28
Add example module using `dusk-merkle`


ureeves:
#### Summary
To ensure we can represent Merkle trees in a module, we should create an example that uses the `dusk-merkle` crate to store a custom Merkle tree.

#### Relevant Context
The transfer contract uses a Merkle tree to store the hashes of a Phoenix note.

#### Possible Solution (optional)
`dusk-merkle` is `no-std` so it should be pretty straightforward to add a module containing a tree as its state.

PR #198, opened 2023-04-28
Add example merkle tree module

ureeves:
Included is a test that entails inserting multiple leaves while querying the root. The obtained roots are then all compared together, ensuring they're all different.

Resolves: #197

### DIFF
--- a/modules/Cargo.toml
+++ b/modules/Cargo.toml
@@ -10,6 +10,7 @@ members = [
     "fallible_counter",    
     "fibonacci",
     "host",
+    "merkle",
     "metadata",
     "micro",
     "crossover",
--- a/modules/README.md
+++ b/modules/README.md
@@ -15,6 +15,7 @@ This workspace contains individual module examples. These examples demonstrate v
 - [Fallible counter](fallible_counter/): Example of a counter that can panic if wanted.
 - [Fibonacci](fibonacci/): Fibonacci and in-module recursion example.
 - [Host](host/): Module that performs a simple host call.
+- [Merkle](merkle/): A Merkle tree in an example module.
 - [Metadata](metadata/): Example of module metadata retrieval.
 - [Micro](micro/): Minimal module example.
 - [Spender](spender/): Module testing the gas spending behavior.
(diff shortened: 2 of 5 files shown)


HDauven (approved):
LGTM, would only kindly suggest to update the Modules README with the newly added `Merkle` module.

As a side note, it might be interesting to add test cases for an empty tree root and duplicate insertions, in case we want to enshrine behavior. Not sure if it is worth the bother though.

Example:
```rust
#[test]
pub fn empty_tree_root() -> Result<(), Error> {
    let vm = VM::ephemeral()?;
    let mut session = vm.session(SessionData::builder())?;

    // (measured) minimum points to pass - insertion in a merkle tree is
    // "expensive".
    session.set_point_limit(147456);

    let id = session
        .deploy(module_bytecode!("merkle"), ModuleData::builder(OWNER))?;

    let root: [u8; 32] = session
        .query(id, "root", &())
        .expect("root query should succeed");

    let expected_empty_root: [u8; 32] = [0u8; 32];

    assert_eq!(
        root, expected_empty_root,
        "empty tree root should match expected value"
    );

    Ok(())
}

#[test]
pub fn duplicate_insertion() -> Result<(), Error> {
    let vm = VM::ephemeral()?;
    let mut session = vm.session(SessionData::builder())?;

    // (measured) minimum points to pass - insertion in a merkle tree is
    // "expensive".
    session.set_point_limit(147456);

    let id = session
        .deploy(module_bytecode!("merkle"), ModuleData::builder(OWNER))?;

    let leaves: [u64; 2] = [42, 42];
    let mut roots = [[0u8; 32]; 2];

    roots
        .iter_mut()
        .zip(leaves)
        .enumerate()
        .for_each(|(i, (root, leaf))| {
            session
                .transact::<_, ()>(id, "insert", &(i as u64, leaf))
                .expect("tree insertion should succeed");

            *root = session
                .query(id, "root", &())
                .expect("root query should succeed");
        });

    let root: [u8; 32] = session
        .query(id, "root", &())
        .expect("root query should succeed");

    // All roots are different from each other
    for (i, lr) in roots.iter().enumerate() {
        for rr in roots
            .iter()
            .enumerate()
            .filter_map(|(j, root)| (i != j).then_some(root))
        {
            assert_ne!(lr, rr, "roots should be different");
        }
    }

    Ok(())
}
```

HDauven (approved):

Issue #202, opened 2023-05-03
Moving a `Session` after module instantiation results in dangling pointers


ureeves:
**Describe the bug**
If a module is instantiated within the scope of a `Session` - such as in the case of a `deploy` or a `query/transact` call - and the session is subsequently moved, the next call to a module in the context of that same session will produce a `SIGSEGV`.

This is due to the way in which we extend the lifetime of a `Session` during module instantion in `WrappedInstance::new`:

https://github.com/dusk-network/piecrust/blob/588c501dd2e505c1c02a0541b54769607aeeb68b/piecrust/src/instance.rs#L151-L155

In hindsight this is obvious, since on a move of a `Session` the pointer underneath the reference whose lifetime is extended will point to invalid memory.

**To Reproduce**

```rust
/// Instantiate a session and perform a query on an existing module.
fn make_session(vm: &VM) -> Session {
    let mut session = vm.session(/*session data*/);
    session.query::<_, ()>(module_id, "some_call", &()).unwrap();
    session
}

/// Instantiate the `VM`, call `make_session()` and then perform another 
/// call on that same session.
fn will_segv() {
    let vm = VM::ephemeral();
    let mut session = make_session(&vm);
    // This call will SIGSEGV
    session.query::<_, ()>(module_id, "some_call", &()).unwrap();
}
```

**Expected behaviour**
It shouldn't segfault.

**Additional context**
This is a typical usage pattern in `rusk` and should really be fixed. It might be possible to fix this by ensuring a `Session` is [pinned](https://doc.rust-lang.org/stable/std/pin/).

PR #203, opened 2023-05-04
Fix dangling pointer after session move

ureeves:
This fix introduces a `SessionInner`, containing the previous contents of the `Session`, and changes the fields of a session to contain a `&'static mut SessionInner` and a reference count.

We can then internally clone a session, which aliases the reference to the `SessionInner`, allowing us to place a session directly in the `Env`.

It is then ensured that a session frees the `SessionInner` on `Drop`ping when the reference count reaches zero.

Resolves: #202

### DIFF
--- a/piecrust/src/session.rs
+++ b/piecrust/src/session.rs
@@ -49,16 +49,44 @@ unsafe impl Sync for Session {}
 /// then be [`commit`]ed to, or discarded by simply allowing the session to
 /// drop.
 ///
-/// New modules are to be `deploy`ed in the context of a session. Metadata
-/// queryable by modules can be set using [`set_meta`].
+/// New modules are to be `deploy`ed in the context of a session.
 ///
 /// [`VM`]: crate::VM
 /// [`queried`]: Session::query
 /// [`transacted`]: Session::transact
 /// [`commit`]: Session::commit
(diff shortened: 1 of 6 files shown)


Daksh14:
Hey LGTM! I would say we can use RefCell to count references instead of manually doing it yourself.

piecrust/src/session.rs
```diff
@@ -49,16 +49,44 @@ unsafe impl Sync for Session {}
...
+/// New modules are to be `deploy`ed in the context of a session.
 ///
 /// [`VM`]: crate::VM
 /// [`queried`]: Session::query
 /// [`transacted`]: Session::transact
 /// [`commit`]: Session::commit
-/// [`set_meta`]: Session::set_meta
 #[derive(Debug)]
 pub struct Session {
+    inner: &'static mut SessionInner,
```
HDauven, line 60: Why not make inner `Rc<RefCell<SessionInner` and let Rust handle the ref count?

On the consuming side, session would expose `borrow()` and `borrow_mut`, or is that not granular enough for this situation?

ureeves: Take a look now and tell me if it makes more sense.

HDauven (approved):

Issue #201, opened 2023-05-03
Owner should of arbitrary length or as large as a BLS public keys


ureeves:
#### Summary
Currently the owner of a contract is defined to be `[u8; 32]`. This should change to be either as large as a `bls12_381_sign::PublicKey`, or perhaps even arbitrary length with `Vec<u8>`.

This is required to use cryptographic keys as owners of contracts, allowing them to, in turn, be used to represent wallets.  

#### Possible solution design or implementation
A simple exchange from `owner: [u8; 32]` to `Vec<u8>` or `[u8; PublicKey::SIZE]` should suffice.

#### Additional context
This is important in the context of the [fee market](https://github.com/dusk-network/rusk/issues/880).

PR #204, opened 2023-05-05
Change `owner` type to `[u8; N]`

ureeves:
This allows the consumer of the `piecrust` crate to define how large
they want the owner to be when offered to a module. Internally, we
handle the owner as a `Vec<u8>`, however it gets passed to a module as
a `[u8; N]`.

Resolves: #201

### DIFF
--- a/piecrust-uplink/src/state.rs
+++ b/piecrust-uplink/src/state.rs
@@ -268,10 +268,10 @@ pub fn height() -> u64 {
 }
 
 /// Return the current module's owner.
-pub fn owner() -> [u8; 32] {
+pub fn owner<const N: usize>() -> [u8; N] {
     let len = unsafe { ext::owner() } as usize;
     with_arg_buf(|buf| {
-        let ret = unsafe { archived_root::<[u8; 32]>(&buf[..len]) };
+        let ret = unsafe { archived_root::<[u8; N]>(&buf[..len]) };
         ret.deserialize(&mut Infallible).expect("Infallible")
     })
 }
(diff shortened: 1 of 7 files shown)


piecrust-uplink/src/state.rs
```diff
@@ -268,10 +268,10 @@ pub fn height() -> u64 {
 }
 
 /// Return the current module's owner.
```
HDauven, line 270: Shouldn't we mention the reasoning here for why we allow N size and why it is even possible to do so in Dusk modules? Given variable size/self defined keys are very uncommon within this context in other blockchains.

HDauven (changes requested):
LGTM but do see my comment pls.

piecrust-uplink/src/state.rs
ureeves: We actually technically are not allowing "variable size" keys at all. We're just allowing the consumer of `piecrust` to define the size of those keys themselves, and possibly create a wrapper around `piecrust_uplink::owner()` returning whatever key type they might want to use. I'm also trying very hard to have piecrust not use cryptography whatsoever - in fact even the concept of "keys" is not present here.

Downstream, in `rusk_abi`, we should provide a function wrapping around the uplink's `owner` function, returning something like a `dusk-bls12_381_sign::PublicKey`, and we would guarantee in `rusk` that this is always set to a valid key.

HDauven: > 

Ah yes. I get what you mean now. Resolving this.

herr-seppia (approved):
LGTM

PR #205, opened 2023-05-09
Remove `query/transact` in favor of `call`

### LINKED ISSUE #136: Semantics of `query` and `transact` are equivalent for modules
#### Summary
The `piecrust_uplink::query` and `piecrust_uplink::State::transact` functions are semantically equivalent. Regardless of whether the module writer decides to use `piecrust_uplink::wrap_query` or `piecrust_uplink::wrap_transact` to wrap their exposed functions the result is the same, and is callable by either a `query` or a `transact` and both will produce the same results.

I propose the distinction between `query` and `transact` be eliminated for the contract writer, replacing it with a `call` function, implementing equivalent functionality.

The solution would be to rename `piecrust_uplink::query` to `piecrust_uplink::call` and remove `piecrust_uplink::State::transact`. Also, the helpers included - `wrap_query` and `wrap_transaction` - should be replace by a single `wrap_call`. 

#### Additional context
It is important to note that what is said above **does not apply** to `Session::query` and `Session::transact` since the semantics are different there. Queries don't mutate the state, while transactions do.

ureeves:
Resolves #136

### DIFF
--- a/modules/box/src/lib.rs
+++ b/modules/box/src/lib.rs
@@ -44,11 +44,11 @@ impl Boxen {
 /// Expose `Boxen::set()` to the host
 #[no_mangle]
 unsafe fn set(arg_len: u32) -> u32 {
-    uplink::wrap_transaction(arg_len, |to| STATE.set(to))
+    uplink::wrap_call(arg_len, |to| STATE.set(to))
 }
 
 /// Expose `Boxen::get()` to the host
 #[no_mangle]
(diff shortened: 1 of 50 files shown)


moCello (approved):
Great change!

HDauven (approved):

fed-franz (approved):
LGTM

Very happy to see this change

PR #208, opened 2023-05-15
Remove `module` nomenclature in favor of `contract`

### LINKED ISSUE #207: Rename "module" to "contract"
#### Summary
We should replace the term `module` in favor of  `contract`. While each contract is *technically* a WASM module, we should make it clear that we are speaking of smart contracts. 

#### Possible Solution
Search for all instances of the term `module` and replace them with `contract` when appropriate.

ureeves:
Resolves #207


HDauven (approved):
LGTM

1+ for bringing the nomenclature in line with what most "Web3" devs would expect.

PR #210, opened 2023-05-15
Remove `piecrust_uplink::State` struct

### LINKED ISSUE #209: Remove `piecrust_uplink::State` struct
#### Summary
The `piecrust_uplink::State` struct is not providing any additional functionality apart from functioning as a container for a contract's state.

Remove `State` struct and move the `State::emit` code to the same level as `call`.

ureeves:
Resolves #209

### DIFF
--- a/modules/box/src/lib.rs
+++ b/modules/box/src/lib.rs
@@ -14,7 +14,6 @@ extern crate alloc;
 use alloc::boxed::Box;
 
 use piecrust_uplink as uplink;
-use uplink::State;
 
 /// Struct that describes the state of the box module
 // One Box, many `Boxen`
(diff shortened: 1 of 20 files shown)


ureeves:
> What was the rationale for having this State struct?

It was originally reasoned that the `State` struct would allow us to leverage the borrow checker to prevent aliasing of the argument buffer. However, we operate in a regime where we can't really guarantee this anyway - seeing that an inter-contract call may result in any part of the caller contract's memory being touched.

Think of the combo with the genesis contracts:

```
          1        2
transfer -> stake -> transfer
```

Call `2` will write to the argument buffer of the transfer contract, effectively making any reference to it obtained before call `1` invalid. `State` was supposed to help with this by providing useful guarantees.

In the end we can already discourage direct reference to the argument buffer by simply not exposing it, but we will need to keep this in mind if we ever wish to use different languages with the VM.

PR #217, opened 2023-06-01
Handle contract-passed points_limit

### LINKED ISSUE #216: Allow callers to specify the points in a call
#### Summary
Rather than forcing the 93% point passing rule - meaning the called contract gets the 93% of the points left in the caller's execution - the caller should be able to specify the number of points they're willing to provide.

It is possible to introduce a number passed with the `call` extern representing the number of points available to the called contract. This would mean the following lines:

https://github.com/dusk-network/piecrust/blob/eb9e90a52b387011e6e7efc18e96d7a11de3e698/piecrust-uplink/src/state.rs#L44-L49

should become:

```rust
pub(crate) fn c( 
    mod_id_ofs: *const u8, 
    name: *const u8, 
    name_len: u32, 
    arg_len: u32,
    point_limit: u64, // new argument
) -> i32;
```

Then the changes should be handled by the VM appropriately.

#### Additional context
This change is important to achieve parity with `rusk-vm`. Somehow this feature was not properly ported when making bootstrapping `piecrust`. Fortunately it is possible to do so now.

ureeves:
Allows for the point limit to be passed by a contract when performing a call.

Resolves #216

### DIFF
--- a/contracts/callcenter/src/lib.rs
+++ b/contracts/callcenter/src/lib.rs
@@ -9,6 +9,7 @@
 #![no_std]
 
 use piecrust_uplink as uplink;
+use piecrust_uplink::call_with_limit;
 use uplink::{wrap_call, ContractError, ContractId, RawCall, RawResult};
 
 /// Struct that describes the state of the Callcenter contract
@@ -79,6 +80,18 @@ impl Callcenter {
             false => Ok(caller == self_id),
         }
     }
+
+    /// Calls the `spend` function of the `contract` with no arguments, and the
+    /// given `points_limit`, assuming the called function returns `()`. It will
+    /// then return the call's result itself.
+    pub fn call_spend_with_limit(
+        &self,
+        contract: ContractId,
+        points_limit: u64,
+    ) -> Result<(), ContractError> {
+        call_with_limit(contract, "spend", &(), points_limit)?;
+        Ok(())
+    }
 }
 
 /// Expose `Callcenter::query_counter()` to the host
@@ -103,6 +116,14 @@ unsafe fn calling_self(arg_len: u32) -> u32 {
 #[no_mangle]
 unsafe fn call_self(arg_len: u32) -> u32 {
     wrap_call(arg_len, |_: ()| STATE.call_self())
+}
+
+/// Expose `Callcenter::call_spend_with_limit` to the host
+#[no_mangle]
+unsafe fn call_spend_with_limit(arg_len: u32) -> u32 {
+    wrap_call(arg_len, |(contract, points_limit)| {
+        STATE.call_spend_with_limit(contract, points_limit)
+    })
 }
 
 /// Expose `Callcenter::return_self_id()` to the host
--- a/contracts/spender/src/lib.rs
+++ b/contracts/spender/src/lib.rs
@@ -50,6 +50,11 @@ impl Spender {
             false => (limit, spent_before, 0, 0, 0),
         }
     }
+
+    /// Spend all points that are given to the contract.
+    pub fn spend() {
+        panic!("I like spending");
+    }
 }
 
 /// Expose `Spender::get_limit_and_spent()` to the host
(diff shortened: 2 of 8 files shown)


piecrust-uplink/src/state.rs
```diff
@@ -134,10 +136,38 @@ where
...
+}
+
+/// Calls a `contract`'s `fn_name` function with the given argument `fn_arg`,
+/// allowing it to spend the given `points_limit`.
+///
+/// A points limit of `0` is equivalent to using [`call`], and will use the
+/// default behavior - i.e. the called contract gets `93%` of the remaining
+/// points.
+///
+/// If the points limit given is not under the remaining amount, the default
```
moCello, line 164: Maybe better say "If the points limit given is *above or equal* the remaining amount, ..."

piecrust-uplink/src/state.rs
```diff
@@ -134,10 +136,38 @@ where
...
+/// behavior will be used instead.
```
moCello, line 165: Is it possible to give a point limit that is above the 93%?

piecrust-uplink/src/state.rs
```diff
@@ -176,20 +205,48 @@ where
...
+}
+
+/// Calls a `contract` with the given [`RawCall`] allowing it to spend the given
+/// `points_limit`.
+///
+/// A point limit of `0` is equivalent to using [`call_raw`], and will use the
+/// default behavior - i.e. the called contract gets `93%` of the remaining
+/// points.
+///
+/// If the points limit given is not under the remaining amount, the default
```
moCello, line 227: see comment above

contracts/callcenter/src/lib.rs
```diff
@@ -79,6 +80,18 @@ impl Callcenter {
             false => Ok(caller == self_id),
         }
     }
+
+    /// Calls the `spend` function of the `contract` with no arguments, and the
+    /// given `points_limit`, assuming the function does not return. It will
+    /// then return that call's result.
+    pub fn call_spend_with_limit(
```
moCello, line 87: So this is spending all gas-points up until the given `points_limit`? What I don't get is what the sentence "..., assuming the function does not return." -> which function, `spend`? `call_spend_with_limit`
"It will then return that call's result" -> which call's? And what is the return?

piecrust/src/imports.rs
```diff
@@ -64,7 +65,12 @@ fn c(
     let caller_remaining = instance
         .get_remaining_points()
         .expect("there should be points remaining");
-    let callee_limit = caller_remaining * POINT_PASS_PCT / 100;
+
+    let callee_limit = if points_limit > 0 && points_limit < caller_remaining {
+        points_limit
+    } else {
+        caller_remaining * POINT_PASS_PCT / 100
+    };
```
moCello, line 73: So it is possible that the `point_limit` is larger than the 93% of the remaining points. Could that be an issue? I guess not but it might be worth to have an extra look at this.

moCello (approved):
LGTM, Some questions regarding the callcenter but I assume these are simply unclear in my own understanding and not actually in the comments

piecrust-uplink/src/state.rs
ureeves: Positive is better than negative. Done.

contracts/callcenter/src/lib.rs
ureeves: I suppose I meant a `void/()` return on the `spend` call. I'll update to reflect this better.

piecrust-uplink/src/state.rs
ureeves: Yes - answer underneath.

piecrust/src/imports.rs
ureeves: The idea is that technically it is up to the contract writer to choose the value appropriately. If it really wants to allow the called contract to consume most/all of its gas it should be allowed to do so.

contracts/spender/src/lib.rs
```diff
@@ -50,6 +50,11 @@ impl Spender {
             false => (limit, spent_before, 0, 0, 0),
         }
     }
+
+    /// Spend all points that are given to the contract.
+    pub fn spend() {
+        panic!("I like spending");
```
moCello, line 56: :laughing:

moCello:
Nice PR, well documented!

Issue #218, opened 2023-06-06
Allow raw calls from the host


ureeves:
#### Summary
To allow for the host to perform calls without knowing the type of the argument, `piecrust` should expose `call_raw` functionality.

#### Possible solution design or implementation
Raw calls are simply calls that emplace directly to the argument buffer, as opposed to first serializing and then de-serializing. As such, it should be a simple copy of `call`, without the serialization functionality.

#### Additional context
This functionality is important to be able to perform arbitrary queries - i.e. queries of which the type is not defined by the host.

PR #219, opened 2023-06-06
Add `Session::call_raw`

ureeves:
This will allow us to defer (de)serialization to the host - or whomever the host requires to do so. It is necessary in order to support generic queries over the network.

Resolves: #218


HDauven (approved):
LGTM

fed-franz (approved):

PR #226, opened 2023-06-16
Add example contract written in C

ureeves:
An experiment in using different languages to target the `piecrust` VM. It resulted in a contract, written in C, that calls the counter contract to increment the counter, and then calls it again to read the counter value. It then proceeds to return the value back to the caller.

As a consequence of the experiment, it was discovered that we don't really require the `__heap_base` export and as such it is removed in this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ contracts: ## Build example contracts
 	  -Z build-std=core,alloc,panic_abort \
 	  -Z build-std-features=panic_immediate_abort \
 	  --target wasm32-unknown-unknown
+	@contracts/c-example/build.sh
 	@mkdir -p target/stripped
 	@find target/wasm32-unknown-unknown/release -maxdepth 1 -name "*.wasm" \
 	    | xargs -I % basename % \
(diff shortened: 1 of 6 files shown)


ureeves:
I will need to add the appropriate build dependencies for the contract - such as `clang` - to the builder machine. Feel free to run it locally while I do.

herr-seppia:
> I will need to add the appropriate build dependencies for the contract - such as `clang` - to the builder machine. Feel free to run it locally while I do.

At your disposal

moCello (approved):
Pretty cool! Basically this the example shows that we can add contracts in any language that can be compiled to WASM

HDauven:
> Pretty cool! Basically this the example that we can add contracts in any language that can be compiled to WASM

From what I understand about Piecrust, a WASM contract only needs access to the arg buffer (for reading and writing to the host) and the host functions.

In essence, we only need to expose helper functions for any language we want to target to ease the developer experience.

I'd love to see/try to make a contract with Python as well, to see how the managed-unmanaged portion is handled in the WASM conversion.

robert-haas:
> I'd love to see/try to make a contract with Python as well, to see how the managed-unmanaged portion is handled in the WASM conversion.

A Python example would be awesome!

Could Go also be an option or is it adding too much baggage in its output?
https://go.dev/doc/go1.11#wasm
https://github.com/golang/go/wiki/WebAssembly

PR #227, opened 2023-06-19
Implement `Display` for `ContractError`

ureeves:


piecrust-uplink/src/error.rs
```diff
@@ -52,3 +54,13 @@ impl From<ContractError> for i32 {
         }
     }
 }
+
+impl Display for ContractError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        match self {
+            ContractError::PANIC => write!(f, "CONTRACT PANIC"),
```
Daksh14, line 61: nitpick: Any reason why the variants are in screaming case? Rust/clippy will complain about that

Daksh14 (approved):
LGTM, just one nitpick

piecrust-uplink/src/error.rs
ureeves: The CI contains clippy checks. No complaints from it so far.

PR #231, opened 2023-06-23
Enrich event data and move to `uplink`

ureeves:
The events are implemented with data according to the [rusk wiki](https://github.com/dusk-network/rusk/wiki/Rusk-Event-System),and moved to `piecrust-uplink` to be made usable to both contracts and host alike.

Currently contracts are only allowed to emit events with themselves as targets, and without specifying `cancelable` or `capturable`.


ZER0 (approved):
LGTM!

PR #234, opened 2023-06-28
Re-export the contents of `piecrust-uplink` in `piecrust`

ureeves:
This ensures that `piecrust` is the only crate required for defining and interacting with a VM, while also allowing for any compatible version of `piecrust-uplink` to be used.


moCello (approved):
LGTM

PR #241, opened 2023-07-10
Introduce call receipts and limit on calls

### LINKED ISSUE #181: Better structure and data encapsulation

ureeves:
We make these changes to better structure the way in which we consume data about calls. Instead of object orientation - where we first set the point limit and only then call - we take a more stack-frame-like approach, where all data about a call is delivered with the call itself.

Resolves #181

### DIFF
--- a/piecrust/src/session.rs
+++ b/piecrust/src/session.rs
@@ -690,12 +690,6 @@ impl Session {
         self.inner.contract_session.clear_contracts();
         self.inner.call_count = 0;
 
-        // TODO Figure out how to handle metadata and point limit.
-        //      It is important to preserve their value per call.
-        //      Right now it probably won't bite us, since we're using it
-        //      "properly", and not setting these pieces of data during the
-        //      session, but only at the beginning.
-
         // This will always be set by the loop, so this one will never be
         // returned.
         let mut res = Ok((vec![], 0));


piecrust/src/session.rs
```diff
@@ -774,6 +762,42 @@ impl Session {
     }
 }
 
+/// The receipt given for a call execution using one of either [`call`] or
+/// [`call_raw`].
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct CallReceipt<T> {
```
HDauven, line 768: For future reference, if all ICCs were to issue a `CallReceipt`, it would make sense to add a `ContractId` to this structure. 

Searching for metadata on events for a given contract would be a bit easier.

HDauven (approved):
LGTM

I like this change a lot, makes reasoning about gas flows through ICC easier.

Added a note to the `CallReceipt` structure for future reference. Not 100% sure if we should do something with it, but time will tell.

herr-seppia (approved):
LGTM

PR #242, opened 2023-07-11
Expand error handling

ureeves:
It is important to handle errors properly, especially when these errors occur due to possibly malicious entities - be it contract or callers - trying to attack a long lived instance of the virtual machine.

This PR does this by handling errors appropriately, in particular in the WASM imports.

### DIFF
--- a/piecrust/src/imports.rs
+++ b/piecrust/src/imports.rs
@@ -96,8 +96,7 @@ fn c(
         .with_memory_mut::<_, Result<_, Error>>(|memory| {
             let name = core::str::from_utf8(
                 &memory[name_ofs as usize..][..name_len as usize],
-            )
-            .expect("TODO error handling");
+            )?;
 
             let arg_buf = &memory[argbuf_ofs..][..ARGBUF_LEN];
 
@@ -143,7 +142,7 @@ fn hq(
     name_ofs: i32,
     name_len: u32,
     arg_len: u32,
-) -> u32 {
+) -> Result<u32, Error> {
     let env = fenv.data_mut();
 
     let instance = env.self_instance();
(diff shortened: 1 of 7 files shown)


piecrust/src/session.rs
```diff
@@ -554,11 +555,11 @@ impl Session {
 
         let mut serializer =
             StandardBufSerializer::new(ser, scratch, Infallible);
-        serializer.serialize_value(value).expect("Infallible");
+        serializer.serialize_value(value)?;
 
         let pos = serializer.pos();
 
-        buf[..pos].to_vec()
+        Ok(buf[..pos].to_vec())
```
herr-seppia, line 562: Ditto

herr-seppia (approved):
LGTM

Not sure if there is the need of bounds check

piecrust/src/imports.rs
```diff
@@ -174,28 +176,24 @@ fn hd(mut fenv: FunctionEnvMut<Env>, name_ofs: i32, name_len: u32) -> u32 {
...
-                buf[..data.len()].copy_from_slice(&data);
-            });
+    let data = env.meta(&name).unwrap_or_default();
 
-            data.len() as u32
-        }
-        _ => 0u32,
-    }
+    instance.with_arg_buffer(|buf| {
+        buf[..data.len()].copy_from_slice(&data);
```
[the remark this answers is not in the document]
ureeves: For this one this is unnecessary. I pushed a commit to make this a little clearer. Metadata items are always `<=` the arg buffer length.

piecrust/src/session.rs
ureeves: If the type is too large to fit in `MAX_META_SIZE` then this will error out anyway. Fortunately

PR #244, opened 2023-07-17
Introduce `feed` to allow contracts to report state

### LINKED ISSUE #243: Allow for contracts to report their data in a single call
#### Summary
Contracts, due to the argument buffer being limited in size, are limited to passing 64KiB per call to the host. We would like to introduce a way for contracts to report their data in a more streamlined way, without leaving the boundaries of a single execution.

It should be possible to "stream" data from the contract by allowing the caller to pass a channel into the contract, and have it emit events that subsequently get passed through this stream.

#### Additional context
The way the data is passed through and the capabilities offered to the contract should be carefully considered. It is probably better to not offer new capabilities to the contract, to avoid introducing any vulnerability.

ureeves:
The `feed` extern allows contracts to stream their state to the host through a channel, allowing for them to effectively report on the entirety of their state - or portions of it - in a single call.

Previously, contracts could only pass data through to the host by returning, or emitting events. This, coupled with the `ARGBUF_LEN` limit, meant that the contract could only send 64KiB through to the host on each execution. This PR removes that restriction.

Resolves #243


piecrust-uplink/src/abi/state.rs
```diff
@@ -295,3 +297,26 @@ where
         unsafe { ext::emit(topic_ptr, topic_len, arg_len) }
     });
 }
+
+/// Feeds the host with data.
+///
+/// This is only allowed to be called in the context of a `feed_call`, and
+/// will error out otherwise. It is meant for contracts to be able to report
```
HDauven, line 304: Should we put some smart constraints on this on the Rusk side? I can imagine this can be used to DDoS nodes.

piecrust/src/session.rs
```diff
@@ -362,6 +360,53 @@ impl Session {
         })
     }
 
+    /// Execute a *feeder* call on the current state state of this session.
```
HDauven, line 363: Duplicate `state`

HDauven (approved):
LGTM

piecrust-uplink/src/abi/state.rs
ureeves: Definitely. Especially when offering these as a service to the wider web, at least a rate limit will be required. I'm unsure as to whether the appropriate place for this kind of defensive measure should be at the infrastructure level, or at the level of rusk itself. We'll see.

piecrust/src/session.rs
ureeves: :man_facepalming:

PR #247, opened 2023-07-26
Fix downstream SIGBUS by adapting memory growth mechanism

ureeves:
The representation of memory is changed to make growth a cheap operation, as opposed to extremely expensive. Previously the memory was copied onto a temporary file with the new size, now growth happens onto an anonymous region of memory beyond the contents of the file, but still contiguous.

The diffing algorithm is slightly adapted to these changes, and made more performant.


piecrust/src/store/memory.rs
```diff
@@ -186,3 +181,288 @@ impl From<Memory> for VMMemory {
         VMMemory(Box::new(memory))
     }
 }
+
+/// The memory used by a contract.
+///
+/// It consists of one large mmap that doesn't reserve swap space - see
+/// [MAP_NORESERVE]. The memory region mapped will always be the maximum
+/// possible, since virtual address space is cheap, but the writable area is
+/// controlled bu `len`.
```
miloszm, line 190: nit: `bu` should be `by`

piecrust/src/store/memory.rs
```diff
@@ -186,3 +181,288 @@ impl From<Memory> for VMMemory {
...
+    /// the file is mmapped copy-on-write, reserving a region as large as
+    /// `MAX_MEM_SIZE`. In the second step an overlapping region is mapped to
+    /// physical memory, at an offset of file length, using `MAP_FIXED`. This
+    /// ensures the memory is free to grow beyond the boundaries of the file
+    /// into physical memory, in one contiguous memory region.
+    ///
+    /// # Errors
+    /// If the file already exists, and it has a length that is not a multiple
+    /// of WASM_PAGE_SIZE, the function will error.
+    pub fn map<P: AsRef<Path>>(file: P) -> io::Result<Self> {
```
miloszm, line 263: nit: this `pub` could go away

piecrust/src/store/memory.rs
```diff
@@ -186,3 +181,288 @@ impl From<Memory> for VMMemory {
...
+}
+
+impl MemoryMmap {
+    /// Creates a new memory, backed by the system's physical memory.
+    ///
+    /// Memories created like this always have `MIN_MEM_SIZE`, unless
+    /// [`grow_by`] is called.
+    ///
+    /// [`grow_by`]: MemoryMap::grow_by
+    pub fn new() -> io::Result<Self> {
```
miloszm, line 235: nit: this `pub` could go away

miloszm (approved):
3 nits but otherwise LGTM

piecrust/src/store/diff.rs
```diff
@@ -10,40 +10,72 @@ use std::io::{Read, Write};
...
-/// larger than it's current length.
+/// Applies the given patch on top of the new memory, using the old one as a
+/// base for the patch.
+///
+/// First, the growth delta is read from the patch, and if it's larger than zero
+/// both memories are grown. Then the patch is applied on top on the new memory,
+/// stemming from the contents of the old.
+///
+/// # Important
+/// The old memory should be discarded (dropped) after this operation.
```
herr-seppia, line 62: If it's so important, did you consider to take the ownership of the old Memory?

piecrust/src/store/memory.rs
```diff
@@ -186,3 +181,288 @@ impl From<Memory> for VMMemory {
...
+        let new_len = self.len + delta_bytes;
+
+        if new_len > MAX_MEM_SIZE {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                format!("Tried growing past the maximum number of bytes: {MAX_MEM_SIZE}"),
+            ));
+        }
+
+        self.len = new_len;
```
herr-seppia, line 323: Does it worth to check if the len is still a multiple of wasm page?

piecrust/src/store/memory.rs
```diff
@@ -186,3 +181,288 @@ impl From<Memory> for VMMemory {
...
+    }
+
+    fn size(&self) -> Pages {
+        let pages = self.len / WASM_PAGE_SIZE - 1; // there is one guard page
+        Pages(pages as u32)
+    }
+
+    fn style(&self) -> MemoryStyle {
+        MemoryStyle::Static {
+            bound: Pages(WASM_MAX_PAGES),
```
herr-seppia, line 370: Should be "-1" due to guard page?

herr-seppia (approved):
LGTM

piecrust/src/store/diff.rs
```diff
@@ -10,40 +10,72 @@ use std::io::{Read, Write};
...
-    Bsdiff::new(old, new).compare(writer)
+    debug_assert!(
+        new_len >= old_len,
+        "Length of memories should strictly increase"
+    );
+
+    let delta_len = new_len - old_len;
+    let delta_len_bytes = delta_len.to_le_bytes();
+
+    writer.write_all(&delta_len_bytes)?;
```
fed-franz, line 43: does it make sense to do this is delta_len is 0 ?

piecrust/src/store/memory.rs
ureeves: The WASM spec says that bound is *all mapped pages*. That's why I used the full memory size here.

piecrust/src/store/diff.rs
ureeves: We need to know in `patch` if we need to grow the memory or not, so either we put the actual length there, like before, or a diff length.

piecrust/src/store/memory.rs
ureeves: We could do this as a debug assertion, but I wouldn't waste runtime on release mode.

piecrust/src/store/diff.rs
ureeves: Omg, you're right :man_facepalming:

ureeves: Well partially. The memory is just an `Arc` to an inner structure. This `Arc` should be deleted, and not used again.

herr-seppia: Damn, I was almost there

PR #251, opened 2023-08-07
Add contract validation

### LINKED ISSUE #157: Validate contract bytecode
#### Summary
The bytecode for a contract should be validated as much as possible for compliance with the VMs ABI. This is necessary to prevent the deployment of clearly faulty contracts.

There are some expects already when a contract is instantiated. These function as placeholders for the behavior that is expected when a contract does not comply with the ABI.

#### Additional context
Contracts should really always comply with the ABI, but this can also not be validated in all circumstances. The implementer should justify the choices made in this regard.

ureeves:
Contract validation is performed during deployment of the contract, to ensure no invalid contracts ever "hit" the state. The following rules are employed:

- Argument buffer fits in the memory
- There is only one memory, and it's called "memory
- Every exported function matches the `f: i32 -> i32` calling convention

Resolves #157

### DIFF
--- a/contracts/Cargo.toml
+++ b/contracts/Cargo.toml
@@ -11,6 +11,7 @@ members = [
     "feeder",
     "fibonacci",
     "host",
+    "invalid",
     "merkle",
     "metadata",
     "micro",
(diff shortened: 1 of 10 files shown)


herr-seppia (approved):
LGTM

Issue #254, opened 2023-08-10
Remove re-execution by leveraging marking strategy


ureeves:
#### Summary
When an error is encountered during contract execution, the changes to its state due to this execution have to be reverted. Currently, we employ a re-execution strategy, where we mark the given call as erroring, reload memories, and execute up until the point of error. We then proceed to return the error to the caller - be it another contract or the host.

This is an extremely expensive strategy, and can be made better by leveraging the solution of https://github.com/dusk-network/piecrust/issues/253.

#### Possible solution design or implementation
Before each call to the contract, we can copy the current set of dirty pages onto memory - effectively executing a snapshot of the current state - and, when an error is detected, discard the modifications made during the context of said call. If the call does not error, we can simply discard the snapshot and continue.

Issue #253, opened 2023-08-10
Remove diffing by employing a marking strategy


ureeves:
#### Summary
When we are in the process of committing a memory, the operation is performed as follows:
```
1. Load fresh copy
2. Compute diff
3. Write compressed diff to disk
```
This is an memory intensive operation, and the compute time scales with the length of the memory. On load the following has to be performed:
```
1. Map initial memory
2. Load diff to physical memory
3. Apply diff to memory
```
This is also a memory intensive operation, since the diffed pages have to be accessed an written to.

Here we propose, instead of computing the diff, marking dirty WebAssembly pages and writing them clean to disk. This results in the following store and load procedures:
```
Load
1. Map dirty pages onto the appropriate regions
Store
1. Write dirty pages to disk
```

The operations then scale with `O(D)` - `D` being the number of dirty pages - instead of with `O(L)` - `L` being the size of the memory.

> 💡 Add a clear and concise description of **the reason why** this needs to be worked on. Focus on the big picture rather than implementation details. What use case does it support? What is the expected outcome?

#### Possible solution design or implementation
If we map the memory region as only `PROT_READ`, any write will produce a `SIGSEGV`. Signals like this can be caught by setting a signal handler using [`sigaction`](https://www.man7.org/linux/man-pages/man2/sigaction.2.html). This signal handler can then set the protections of the page being written to as `PROT_READ | PROT_WRITE` using [`mprotect`](https://man7.org/linux/man-pages/man2/mprotect.2.html), and mark it as dirty.

#### Additional context
It is important to note that this will also allow us to perform "snapshots" of memory in the future, meaning the re-execution strategy can be replaced by a `snap-apply/discard` strategy on the inter-contract call boundary.

PR #255, opened 2023-08-16
Introduce `crumbles` - keeping track of dirt in userspace

ureeves:
`crumbles` is intended to solve the problem of mapping WASM memories onto virtual address space.

It is built to support tracking dirty pages and creating memory snapshots to which it is possible to revert to. This is required if we want to remove re-execution in `piecrust`, as well as allowing for the removal of diffing in the storage process of state commits.

Since it uses system calls that only exist on *nix systems it can only build on Linux and MacOS. The tracking of dirty pages is done slightly differently on Linux and MacOS, and it might be useful to test it actively for different combinations of platform and architecture. It is important to note that using this would not be a regression for `piecrust` since it currently only builds on *nix as well.

See also: #253
See also: #254


fed-franz (approved):
LGTM!

Nice lib and love the name :)

Couple of nits in the comments.

Also: since you define `MEM_SIZE` and `PAGE_SIZE`, it might be worth using those in the description, instead of assuming is 4Gib/64Kib. This would turn useful if you plan to release the lib with those value as input parameters.

crumbles/src/lib.rs
```diff
@@ -0,0 +1,650 @@
...
+
+    /// Snapshot the current state of the memory.
+    ///
+    /// Snapshotting the memory should be done when the user wants to create a
+    /// point in time to which they can revert to. This is useful when they
+    /// want to perform a series of operations, and either [`revert`] back to
+    /// the original or [`apply`] to keep the changes.
+    ///
+    /// # Errors
+    /// If the underlying call to the protect the memory region fails, this
```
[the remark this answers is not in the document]
ureeves: Fixed!

crumbles/src/lib.rs
```diff
@@ -0,0 +1,650 @@
...
+/// set of files, at the given offsets.
+///
+/// It is possible to create snapshots of the memory, which can be used to
+/// revert to a previous state. See [`snap`] for more details.
+///
+/// Writes are tracked at the page level. This functions as follows:
+///
+/// - When a region is first mapped, its permissions are set to read-only,
+///   resulting in a segmentation fault when a write is attempted.
+/// - The segmentation fault is then caught using a signal handler, which
```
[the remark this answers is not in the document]
ureeves: Fixed

crumbles/src/lib.rs
```diff
@@ -0,0 +1,653 @@
...
+    ///
+    /// let mut contents = Vec::new();
+    /// file.read_to_end(&mut contents).unwrap();
+    ///
+    /// let mmap = unsafe { Mmap::with_files(iter::once(Ok((file, 0)))) }.unwrap();
+    /// assert_eq!(mmap[..contents.len()], contents[..]);
+    /// ```
+    pub unsafe fn with_files<I>(files_and_offsets: I) -> io::Result<Self>
+    where
+        I: IntoIterator<Item = io::Result<(File, usize)>>,
```
moCello, line 140: As a slightly pedantic suggestion I would prefer to have the iterator iterate over `(offset, file)` and not `(file, offset)` purely because this is the tuple that is also returned by a call to `enumerate()`.
But feel free to ignore this

crumbles/src/lib.rs
```diff
@@ -0,0 +1,653 @@
...
+            != 0
+        {
+            return Err(io::Error::last_os_error());
+        }
+
+        Ok(())
+    }
+
+    fn last_snapshot(&self) -> &Snapshot {
+        self.snapshots.last().unwrap()
```
moCello, line 512: potentially adding a comment as to why the unwrap is safe (there is always at least one snapshot present)

crumbles/src/lib.rs
```diff
@@ -0,0 +1,653 @@
...
+    }
+
+    fn last_snapshot_mut(&mut self) -> &mut Snapshot {
+        self.snapshots.last_mut().unwrap()
```
moCello, line 516: same here

crumbles/src/lib.rs
```diff
@@ -0,0 +1,653 @@
...
+            sa_restorer: None,
+        };
+        let mut old_act = MaybeUninit::<sigaction>::uninit();
+
+        if libc::sigaction(libc::SIGSEGV, &act, old_act.as_mut_ptr()) != 0 {
+            process::exit(1);
+        }
+
+        // On Apple Silicon for some reason SIGBUS is thrown instead of SIGSEGV.
+        // TODO should investigate properly
```
moCello, line 553: Is this TODO supposed to be a reminder for the future or did you plan to investigate further before releasing?

moCello (approved):
LGTM, just some tinitiny suggestions but nothing major.
Is rustfmt.toml supposed to be deleted though?

crumbles/src/lib.rs
ureeves: Ah, interesting. There would be no need to `.map` and switch the order. Will do.

crumbles/src/lib.rs
ureeves: It's meant to be just a reminder. If a problem occurs on Apple Silicon, we know where to look.

PR #257, opened 2023-08-24
Port `piecrust` to use `crumbles`

### LINKED ISSUE #253: Remove diffing by employing a marking strategy

ureeves:
Proof-of-life for `crumbles` being used as a backend for memory in `piecrust`. This makes it possible for us to remove diffing, in favor of tracking each dirty page, and only writing new pages when they were dirtied. This is a significantly more scalable approach than computing a diff, and is [more flexible](https://github.com/dusk-network/piecrust/issues/254).

Each contract has 2¹⁶ pages, each of which is placed in a Merkle tree. It then becomes possible to construct proofs of inclusion for each page in the state. We might consider exposing this functionality in the future.

Resolves #253


piecrust/src/instance.rs
```diff
@@ -151,13 +151,13 @@ impl WrappedInstance {
         let instance = wasmer::Instance::new(&mut store, &module, &imports)?;
 
         // Ensure there is a global exported named `A`, whose value is in the
-        // first few pages of memory.
```
herr-seppia, line 154: Curiosity: why this has changed?

herr-seppia (approved):
LGTM!!!

Nothing to say except that the removal of the `squash_commit` function is in the wrong commit (the `init_traps()` one)

moCello (approved):
LGTM

piecrust/src/instance.rs
ureeves: The contract *can technically* put it's argument buffer wherever it wants in its address space. Since we consider all space to be allocated - even if its not really - the argument buffer can be anywhere.

ureeves:
> Nothing to say except that the removal of the `squash_commit` function is in the wrong commit (the `init_traps()` one)
 Strange this happened. Will fix

Issue #206, opened 2023-05-15
Allow for querying call tree


ureeves:
#### Summary
Multiple modules may be touched during the context of a call, forming a tree of inter-contract calls from one module to the next. It should be possible to query this tree.

#### Possible solution design or implementation
The tree should be kept by a `Session` and contain information about the points used during a call.

```rust
struct CallTree {
    module_id: ModuleId,
    points_spent: u64,
    calls: Vec<CallTree>,
}
```

#### Additional context
This is important for implementing the new [economic model](https://github.com/dusk-network/rusk/issues/880).

PR #258, opened 2023-08-28
Remove re-execution in favor of micro-snapshots

ureeves:
This removes the re-execution code and leverages `crumbles`' snapshotting features instead. Each time a contract is called, a snapshot is created. If an error - panic or otherwise - occurs, the changes due to the call are reverted. The call tree is kept track of using the call stack, which will be useful when we address [querying the call tree](https://github.com/dusk-network/piecrust/issues/206).

We also fix a small, but significant, problem with crumbles where we were protecting only the first page of memory, as opposed to all of it.

Resolves: #254
See also: #206


herr-seppia (approved):
LGTM

fed-franz (approved):
A bit of documentation (e.g. on the CallTree and the use of crumbles) wouldn't hurt...

ureeves:
> A bit of documentation (e.g. on the CallTree and the use of crumbles) wouldn't hurt...

Very true. I'll place the images I made in the `docs/` folder, and update the memory document to match.

PR #262, opened 2023-09-07
Add `crumbles::Mmap::set_len` and adapt `piecrust` memory behavior

ureeves:
This makes piecrust properly handle memory growth behavior, and allows for `crumbles` to defer to `wasmer` when a heap access violation is detected.


herr-seppia (approved):
LGTM

PR #265, opened 2023-09-11
crumbles: fix file descriptor leak

ureeves:
Small oversight in the usage of `File::into_raw_fd`, leading to files used by crumbles *never being closed*


HDauven (approved):
LGTM, RTFM

PR #267, opened 2023-09-11
Fix contract/host heap access

ureeves:
This likely fixes the downstream [rusk issue](https://github.com/dusk-network/rusk/issues/1033).

### DIFF
--- a/crumbles/src/lib.rs
+++ b/crumbles/src/lib.rs
@@ -42,7 +42,6 @@
 #![deny(clippy::pedantic)]
 
 use std::{
-    cmp,
     collections::BTreeMap,
     fs::File,
     mem::{self, MaybeUninit},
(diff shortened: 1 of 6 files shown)


crumbles/src/lib.rs
```diff
@@ -473,13 +467,6 @@ impl MmapInner {
         let start_addr = self.bytes.as_ptr() as usize;
         let page_offset = page_num * PAGE_SIZE;
 
-        if page_offset >= self.bytes.len() {
```
herr-seppia, line 476: Shouldn't you still check the OoB? Maybe using max_page

ureeves: Rust still checks out of bounds accesses to the slice itself, and panics if OoB occurs.

herr-seppia (approved):
LGTM

PR #269, opened 2023-09-12
Revert memory growth on errors

### LINKED ISSUE #268: Growth is not rolled back on errors
**Describe the bug**
When a given contract errors during execution, the results of its execution is rolled back. The same should happen for the memory growth that happens due to this execution. However, it does not happen.

**To Reproduce**
It is not necessary to do this - by mere inspection, the the `current_len` of the `VMMemoryDefinition` kept in `Memory` is never reduced - only ever increased.

**Expected behaviour**
We should roll back growth on contract errors, in the same way that memory is rolled back.

ureeves:
Resolves #268

### DIFF
--- a/piecrust/CHANGELOG.md
+++ b/piecrust/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fake guard pages are now removed
 
+### Fixed
+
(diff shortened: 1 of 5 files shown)


HDauven:
Could you provide a test pls for this?

ureeves:
> Could you provide a test pls for this?

Done and done!

HDauven (approved):
LGTM

PR #272, opened 2023-09-22
Support panic messages being passed to the host

### LINKED ISSUE #271: Allow for rich panics
#### Summary
It should be possible for contracts to report the reason for a panic to the host.

```rust
panic!("I am panicking here!");
```

It's likely this will require the addition of a something like a `panic` host call - which should be optional for use. It should also be added to the standard Rust `panic_handler`.

#### Additional context
We had a feature like this in `rusk-vm`, but it seems to have fallen to the wayside since. Luckily it should be easy enough to introduce.

ureeves:
Resolves #271


miloszm (approved):
LGTM, it'd be nice to keep the size check in the Makefile

PR #275, opened 2023-09-26
Receive call tree after execution

ureeves:
Adds the call tree to the receipt of a call, allowing the host to understand which contract spent how much gas during the total execution. This tree can be iterated through, yielding `CallTreeElem` which contain various information about the call that was made. This also means code can be written downstream that branches on what each contract spent.

Resolves: #206

### DIFF
--- a/piecrust/src/call_tree.rs
+++ b/piecrust/src/call_tree.rs
@@ -23,7 +23,7 @@ pub struct CallTreeElem {
 pub struct CallTree(Option<*mut CallTreeNode>);
 
 impl CallTree {
-    /// Creates a new empty call tree, starting with at the given contract.
+    /// Creates a new empty call tree, starting with the given contract.
     pub(crate) const fn new() -> Self {
         Self(None)
     }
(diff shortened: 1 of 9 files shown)


piecrust/src/call_tree.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    for child in children {
+        free_tree(child);
+    }
+}
+
+struct CallTreeNode {
+    elem: CallTreeElem,
+    children: Vec<*mut Self>,
+    prev: Option<*mut Self>,
```
moCello, line 174: Wouldn't it make sense to say `parent` instead of `prev`?

piecrust/src/call_tree.rs
```diff
@@ -0,0 +1,248 @@
...
+    pub spent: u64,
+    pub mem_len: usize,
+}
+
+/// The tree of contract calls.
+#[derive(Debug, Default)]
+pub struct CallTree(Option<*mut CallTreeNode>);
+
+impl CallTree {
+    /// Creates a new empty call tree, starting with at the given contract.
```
fed-franz, line 26: with or at ?

piecrust/src/call_tree.rs
```diff
@@ -0,0 +1,248 @@
...
+impl CallTreeNode {
+    fn new(elem: CallTreeElem) -> *mut Self {
+        Box::leak(Box::new(Self {
+            elem,
+            children: Vec::new(),
+            prev: None,
+        }))
+    }
+
+    fn with_prev(elem: CallTreeElem, prev: *mut Self) -> *mut Self {
```
moCello, line 186: same here: I find `with_parent` more descriptive

moCello (approved):
LGTM, just a naming nit

piecrust/src/call_tree.rs
```diff
@@ -0,0 +1,248 @@
...
+    pub(crate) fn update_spent(&mut self, spent: u64) {
+        if let Some(inner) = self.0 {
+            unsafe {
+                (*inner).elem.spent = spent;
+                update_spent(inner);
+            }
+        }
+    }
+
+    /// Returns the `n`th previous element from the counting from the current
```
fed-franz, line 90: I guess the first "from the" is a leftover

fed-franz (approved):
LGTM
Just a couple of typos

piecrust/src/call_tree.rs
```diff
@@ -0,0 +1,248 @@
...
+struct CallTreeIter<'a> {
+    tree: Option<Subtree>,
+    _marker: PhantomData<&'a ()>,
+}
+
+impl<'a> Iterator for CallTreeIter<'a> {
+    type Item = &'a CallTreeElem;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        unsafe {
```
miloszm, line 208: This is quite a long unsafe block, could it be avoided? If not, could you add a comment to it?

piecrust/src/call_tree.rs
ureeves: Agreed. Substituted for `parent`, including all variables and mentions in comments.

piecrust/src/call_tree.rs
ureeves: Ditto

piecrust/src/call_tree.rs
ureeves: Fixed

piecrust/src/call_tree.rs
ureeves: Merging the call stack with the call tree resulted in this. Fixed!

piecrust/src/call_tree.rs
ureeves: Unfortunately the traversal logic is quite heavy on the pointer dereferences, so the long unsafe block is kind of unavoidable. We also have a nodes referring to both their children *and* their parent, so pointers are unavoidable as well.

I did add a comment on why all these dereferences are safe. We ensure that the tree exists - i.e. the nodes the pointers point to exist - by leveraging a lifetime into the tree when the iterator is created.

miloszm (approved):

PR #276, opened 2023-10-01
Return nothing from `owner` and `self_id`

ureeves:
Given that the return is meant to be the length of what is written by the host in the argument buffer, and this length is fixed for both functions, it can safely be removed.

### DIFF
--- a/piecrust-uplink/src/abi/state.rs
+++ b/piecrust-uplink/src/abi/state.rs
@@ -13,7 +13,7 @@ use rkyv::{
 
 use crate::{
     ContractError, ContractId, RawCall, RawResult, StandardBufSerializer,
-    SCRATCH_BUF_BYTES,
+    CONTRACT_ID_BYTES, SCRATCH_BUF_BYTES,
 };
 
 pub mod arg_buf {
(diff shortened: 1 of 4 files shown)


herr-seppia (approved):
LGTM

HDauven (approved):

piecrust-uplink/src/abi/state.rs
```diff
@@ -237,18 +237,18 @@ where
...
     })
 }
 
 /// Return the current contract's id.
 pub fn self_id() -> ContractId {
-    let len = unsafe { ext::self_id() } as usize;
+    unsafe { ext::self_id() };
     let id: [u8; 32] = with_arg_buf(|buf| {
-        let ret = unsafe { archived_root::<[u8; 32]>(&buf[..len]) };
+        let ret = unsafe { archived_root::<[u8; 32]>(&buf[..32]) };
```
moCello, line 251: it might make sense to introduce a `const ID_LEN: usize = 32`

moCello (approved):
LGTM apart from one comment that you are free to ignore

PR #278, opened 2023-10-10
Map each contract page lazily

ureeves:
This PR addresses an important scaling issue currently present with the page mapping mechanism.

Since there exists one file per contract page, and all pages must be accessible by the contract at any time, all files had to be iterated through and mapped to the address space region to be used as the contract's memory - *immediately* on first load. This will become a significant upfront cost for running a contract, especially once it becomes larger.

Here we introduce a mechanism where the pages are loaded on-demand, removing the upfront cost of loading a contract, and trading it in for more memory usage (2 bits per page).


moCello (approved):
LGTM, well done on the comments!

Issue #280, opened 2023-10-11
Build contracts for `wasm64-unknown-unknown`


ureeves:
#### Summary
Given the fact that a contract can currently [only support a maximum of 2^32 bytes](#159) of data, we should allow `piecrust-uplink` to build properly for the `wasm64-unknown-unknown` target. This will enlarge the address space significantly, and then it will become the job of `piecrust` support such large memories.

#### Possible solution design or implementation
Since [`wasm64-unknown-unknown`](https://doc.rust-lang.org/rustc/platform-support/wasm64-unknown-unknown.html) is a tier 3 target, the implementer will have to build the Rust compiler [from source](https://rustc-dev-guide.rust-lang.org/building/how-to-build-and-run.html#what-is-xpy). They will therefore also have to choose how this compiler is disseminated to developers and contract builders alike.

PR #282, opened 2023-10-14
Use Dusk compiler for contracts

ureeves:
The compiler is downloaded and extracted to the target directory, and linked to the toolchains in the rustup directory, making it available to be used anyone in the system.

The compiler is available as under the `+dusk` option in rust tools, and is able to compile at least the `wasm32-unknown-unknown` target. It is released [here](https://github.com/dusk-network/rust/releases/tag/v0.0.0).

See-also: #280


HDauven (approved):
LGTM

PR #285, opened 2023-10-27
Remove unused dependencies and upgrade to latest versions

ureeves:
We take a moment to check for unused dependencies in the crate, and upgrade when applicable.

### DIFF
--- a/piecrust/src/store.rs
+++ b/piecrust/src/store.rs
@@ -157,7 +157,8 @@ fn read_all_commits<P: AsRef<Path>>(
         let entry = entry?;
         if entry.path().is_dir() {
             let commit = read_commit(entry.path())?;
-            commits.insert(*commit.index.root(), commit);
+            let root = *commit.index.root();
+            commits.insert(root, commit);
         }
     }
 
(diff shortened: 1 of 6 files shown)


herr-seppia (commented):
Remove nested use

piecrust/src/store/tree.rs
```diff
@@ -4,7 +4,10 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-use std::collections::{BTreeMap, BTreeSet};
+use std::{
+    cell::Ref,
+    collections::{BTreeMap, BTreeSet},
+};
```
herr-seppia, line 10: Remove this nested use

piecrust/src/store.rs
```diff
@@ -430,7 +431,12 @@ fn write_commit_inner<P: AsRef<Path>>(
             bytecode_dir,
             memory_dir,
             base: base.map(|inner| {
-                let base_root = inner.index.root();
+                let base_root_ref = inner.index.root();
+                let base_root = *base_root_ref;
+
+                // We manually drop the reference to the root hash to allow us
+                // to move out of `inner` when we return.
+                drop(base_root_ref);
```
herr-seppia, line 439: Isn't `let base_root = *inner.index.root();` enough to release the ref to base_root?

herr-seppia (approved):

moCello (approved):
LGTM

piecrust/src/store.rs
ureeves: You're right, fixed.

piecrust/src/store/tree.rs
ureeves: We use it in other parts of the code here in piecrust, and I tend to prefer it, so I think I'll keep it.

PR #288, opened 2023-10-30
Fix loading state from other archs

### LINKED ISSUE #287: Cannot load state form different archs
**Describe the bug**
See also dusk-network/rusk#1043

**Possible solutions**
- Keep compilations outside the state (This can be handled upstream in rusk-recovery, removing the compiled file from the portable state, but we must be sure that piecrust is going to recompile it if it's missing)
- Recompile if the state if it's not build for the current architecture

ureeves:
We do this by at `VM` instantiation time, checking if the module is correct for the native platform. If the module is not correct we compile it and write it back to disk. 

Resolves #287

### DIFF
--- a/piecrust/src/store.rs
+++ b/piecrust/src/store.rs
@@ -20,6 +20,7 @@ use std::sync::mpsc;
 use std::{fs, io, thread};
 
 pub use bytecode::Bytecode;
+use dusk_wasmtime::{Engine, Module};
 pub use memory::{Memory, MAX_MEM_SIZE, PAGE_SIZE};
 pub use metadata::Metadata;
 pub use objectcode::Objectcode;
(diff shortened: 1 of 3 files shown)


piecrust/src/store.rs
```diff
@@ -196,6 +204,37 @@ fn commit_from_dir<P: AsRef<Path>>(dir: P) -> io::Result<Commit> {
             ));
         }
 
+        // TODO ensure that objectcode is valid for this platform, otherwise
```
herr-seppia, line 207: Errant?

herr-seppia (approved):
LGTM

How much overhead is this PR introducing?
Asking because I'm not sure if it worth to add the check as distinct function

piecrust/src/store.rs
ureeves: Yes, fixed.

PR #289, opened 2023-10-31
Implement proofs of page inclusion

### LINKED ISSUE #273: Implement memory page proof of inclusion
#### Summary

With the recent addition of `crumbles` and memory snapshots, it is now possible to implement memory page proof of inclusion. This feature allows clients to validate that a given memory page is part of the state tree for a given contract, for a given block. 

The memory page proof of inclusion functionality streamlines the sync-up process by requesting a "snapshot" of the current state, transferring it to a client and then allowing them to check the validity of the fully synced state by comparing the block hash to that of other nodes. This makes node and client syncing efficient and reliable.

Piecrust needs to publicly expose an API that allows for the generation of the proof of inclusion for a given memory page. 

Due to future node role segregation and an ever moving state, caveats exist around the snapshot of the state given to clients. Proof of inclusions should thus be able to be ran on either historical state, or a given block height that keeps the asynchronous nature of a node client in mind.

ureeves:
Introduces `Session::memory_pages`, returning an iterator over a contract's pages, together with an inclusion proof - expressed in a new type `PageOpening`.

`PageOpening` leverages `dusk-merkle` to allow for verification that a page is indeed in a state with the given root.
    
Resolves #273


HDauven (approved):
LGTM!

miloszm (approved):
LGTM

piecrust/src/store.rs
```diff
@@ -217,20 +237,13 @@ fn commit_from_dir<P: AsRef<Path>>(
...
                 Module::new(engine, bytecode.as_ref()).map_err(|err| {
                     io::Error::new(io::ErrorKind::InvalidData, err)
                 })?;
-            fs::write(
-                objectcode_path,
-                module.serialize().map_err(|err| {
-                    io::Error::new(io::ErrorKind::InvalidData, err)
-                })?,
-            )?;
+            fs::write(objectcode_path, module.serialize())?;
```
moCello, line 246: since you are replacing `objectcode` with `module` maybe also rename the `objectcode_path` variable? Does something like `modulecode_path` make sense?

moCello (approved):
LGTM

piecrust/src/store.rs
ureeves: You're right. I switched the nomenclature to `module` - without "code".

PR #293, opened 2023-11-21
Fix post-error root calculation 

### LINKED ISSUE #296: Inconsistent root after erroring calls
**Describe the bug**
The root is inconsistent after an erroring call.

**To Reproduce**
During the implementation of https://github.com/dusk-network/rusk/pull/1125, a test was added that makes an erroring call, and then subsequently executes another call. Another session is then instantiated that only runs successful calls, which then displays a different root than before.

**Expected behaviour**
Calls should be atomic - if they error nothing should be left behind.

ureeves:
Resolves #296

### DIFF
--- a/contracts/callcenter/src/lib.rs
+++ b/contracts/callcenter/src/lib.rs
@@ -92,6 +92,11 @@ impl Callcenter {
         call_with_limit(contract, "spend", &(), points_limit)?;
         Ok(())
     }
+
+    /// Just panic.
+    pub fn panik(&self) {
+        panic!("panik");
+    }
 }
 
 /// Expose `Callcenter::query_counter()` to the host
@@ -158,4 +163,10 @@ unsafe fn delegate_transaction(arg_len: u32) -> u32 {
     wrap_call(arg_len, |(mod_id, rt): (ContractId, RawCall)| {
         STATE.delegate_transaction(mod_id, rt)
     })
+}
+
+/// Expose `Callcenter::panik()` to the host
+#[no_mangle]
+unsafe fn panik(arg_len: u32) -> u32 {
+    wrap_call(arg_len, |()| STATE.panik())
 }
--- a/piecrust/src/session.rs
+++ b/piecrust/src/session.rs
@@ -5,7 +5,6 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use std::borrow::Cow;
-use std::collections::btree_map::Entry;
 use std::collections::BTreeMap;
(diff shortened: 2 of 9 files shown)


contracts/callcenter/src/lib.rs
```diff
@@ -92,6 +92,11 @@ impl Callcenter {
         call_with_limit(contract, "spend", &(), points_limit)?;
         Ok(())
     }
+
+    /// Just panic.
+    pub fn panik(&self) {
```
moCello, line 97:
```suggestion
    pub fn panic(&self) {
```

contracts/callcenter/src/lib.rs
```diff
@@ -92,6 +92,11 @@ impl Callcenter {
...
+        panic!("panik");
```
moCello, line 98:
```suggestion
        panic!("panic");
```

contracts/callcenter/src/lib.rs
ureeves: But I like to [`panik`](https://i.pinimg.com/originals/bd/ce/25/bdce2580f547211a9bbf43d04a9b50ff.jpg)

contracts/callcenter/src/lib.rs
ureeves: https://github.com/dusk-network/piecrust/pull/293#discussion_r1402114886

piecrust/src/session.rs
```diff
@@ -451,47 +450,35 @@ impl Session {
...
     ///
-    /// If the contract does not exist, or is otherwise not instantiated in this
-    /// session, it will return `None`.
-    pub fn memory_len(&self, contract_id: &ContractId) -> Option<usize> {
-        self.instance(contract_id)
-            .map(|instance| instance.mem_len())
+    /// If the contract does not exist, it will return `None`.
+    pub fn memory_len(
+        &mut self,
+        contract_id: ContractId,
```
moCello, line 456: Is there a reason why this is changed to consume the `contract_id`?

moCello (approved):
LGTM

piecrust/src/session.rs
ureeves: Yes. Previously it used the `WrappedInstance::mem_len` function internally, but now it uses `ContractSession::contract`. The latter takes a `ContractId` rather than a reference to it. It's essentially for the same reason as why we changed the return to be a `Result`.

This is kk though, since `ContractId` is pretty small and `Copy`.

herr-seppia (approved):
LGTM

PR #299, opened 2023-12-05
Fix improper use of mach_ports

ureeves:
Using a dummy engine instantiated with the default configuration resulted in conflicting signal handling on MacOS. This commit fixes this problem, allowing MacOS users to run.


moCello (approved):
LGTM

PR #302, opened 2023-12-05
Add support for most common developer environments

ureeves:
Upgrades the version of the compiler to support the most common developer targets, and broadens the CI to cover those same targets.

This version of the compiler allows us to build contract binaries on most common smart contract developer targets:

- `aarch64-apple-darwin`
- `x86_64-apple-darwin`
- `x86_64-unknown-linux-gnu`

To achieve this we had to rewrite the `c-example` contract to Rust, seemingly defeating its purpose. However it is important to note that was was really being tested was building a contract without resorting to `piecrust-uplink`, so this shouldn't prove to be an issue. This comes at the benefit of removing `llvm` as a developer dependency, leaving a Rust toolchain as the only dependency for supported platforms.


herr-seppia (approved):
LGTM

Issue #301, opened 2023-12-05
Panic information should be available to calling contracts


ureeves:
#### Summary
Contracts making calls to other contracts should receive detailed information about the cause of a contract panic. Presently, when a failure occurs during the execution of a called contract, the calling contract is only informed of basic details, such as the presence of a panic or an interruption due to running out of points. To address this limitation, it is crucial to enhance the reporting mechanism to include more specific information about the failure, considering its significant downstream implications*.

#### Possible solution design or implementation
A comprehensive solution involves modifying the calling convention, specifically `ext::c`, and potentially adjusting or even removing `ContractError`. The goal is to enable the passing of detailed information regarding the cause of failure. The current approach involves returning a positive number (the return length) on successful calls or a negative number on failures. Enhancing this mechanism would involve incorporating additional information about the failure in the argument buffer, mirroring how data is currently passed in the argument buffer.

#### Additional context
*Downstream this means that it becomes impossible for the transfer contract to pass the reason for an inter-contract call failure back to the caller, apart from a simple "this has panicked".

PR #303, opened 2023-12-08
Allow contracts access to panic information

ureeves:
This pull request enables a contract to access the reason behind a panic that occurs during an inter-contract call. It also includes enhancements to the inter-contract call functions, improving the clarity of arguments.

In specific updates, the enum variants of ContractError have been renamed to camel case. The Panic variant now includes a panic message as a String. The host ensures that this string is placed on the callee's argument buffer.

Resolves: #301

### DIFF
--- a/piecrust/src/error.rs
+++ b/piecrust/src/error.rs
@@ -5,7 +5,6 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use std::borrow::Cow;
-use std::convert::Infallible;
 use std::sync::{mpsc, Arc};
 use thiserror::Error;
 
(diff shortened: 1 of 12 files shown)


piecrust/src/error.rs
```diff
@@ -60,9 +59,9 @@ pub enum Error {
     MissingHostData(String),
     #[error("Missing host query: {0}")]
     MissingHostQuery(String),
-    #[error("OutOfPoints")]
+    #[error("Contract OutOfPoints")]
```
herr-seppia, line 62: It's not the contract that goes OutOfPoints, it's the execution

herr-seppia (approved):
LGTM, just a suggestion

moCello (approved):
nice!

piecrust/src/error.rs
ureeves: I was writing a long justification for this, but I realized you're right. However, this also means that the `Error::ContractPanic` should be named `Error::Panic` and the error message adjusted accordingly.

PR #304, opened 2023-12-11
Change `points` terminology to `gas`

ureeves:
There's been a lot of confusion on the difference between `points` and `gas`.

The theory was that we would use different terminology here, and that there would be a one-to-one relationship between gas and points upstream. This would also ensure that the implementer would be careful in translating points to actual value.

However, as the reader will no doubt notice, this constitutes unnecessary cognitive overhead for no tangible benefit. This is due to the fact that a layer of translation between gas and value has to exist regardless of whether the `points/gas` distinction exists or not.

Therefore this PR changes the entire nomenclature of the project to use `gas`, as opposed to `points`, effectively eliminating this overhead.

### DIFF
--- a/contracts/c-example/src/lib.rs
+++ b/contracts/c-example/src/lib.rs
@@ -21,7 +21,7 @@ mod ext {
             fn_name: *const u8,
             fn_name_len: u32,
             fn_arg_len: u32,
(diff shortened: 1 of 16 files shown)


moCello (approved):
Nice

PR #310, opened 2024-01-11
Fix module recompilation

ureeves:
Fixes recompiling the modules on VM instantiation when invalid ones are detected. This can happen in many circumstances, one of which is ingesting state produced under a different architecture or operating system.


herr-seppia (approved):
LGTM

PR #312, opened 2024-01-12
Allow for querying a specific contract's owner

ureeves:
Currently it is only possible for contract's to query for their own owners. This PR makes it possible for contract's to query for any other contract's owner as well.

This is achieved by modifying the `owner` import to accept a pointer to a contract ID in the calling contract's memory. If the passed pointer is non-null, the host reads the contract ID from the contract's memory, otherwise it is assumed to be the calling contract's.

If the contract queried for exists, its owner is emplaced in the calling contract's argument buffer, and 1 is returned, otherwise 0 is returned.

### DIFF
--- a/piecrust/tests/metadata.rs
+++ b/piecrust/tests/metadata.rs
@@ -47,3 +47,75 @@ fn metadata() -> Result<(), Error> {
 
     Ok(())
 }
+
+#[test]
+fn owner_of() -> Result<(), Error> {
+    const EXPECTED_OWNER_0: [u8; 33] = [3u8; 33];
+    const EXPECTED_OWNER_1: [u8; 33] = [4u8; 33];
+
+    const CONTRACT_ID_0: ContractId = ContractId::from_bytes([1; 32]);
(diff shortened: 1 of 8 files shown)


piecrust/tests/metadata.rs
```diff
@@ -47,3 +47,75 @@ fn metadata() -> Result<(), Error> {
...
+            "read_owner_of",
+            &CONTRACT_ID_2,
+            LIMIT,
+        )?
+        .data;
+
+    assert_eq!(
+        owner,
+        None,
+        "The first contract should think that the owner of a non-existing contract in None"
```
fed-franz, line 117:
```suggestion
        "The first contract should think that the owner of a non-existing contract is None"
```

fed-franz (approved):
LGTM
pointed a typo (I think)

moCello (approved):
LGTM

miloszm (approved):

piecrust/tests/metadata.rs
ureeves: Fixed!

Issue #313, opened 2024-01-15
Add ability to migrate contract bytecode


ureeves:
#### Summary
In the interests of allowing the Dusk protocol to be upgradable, and given that part of the protocol is expressed in smart contracts, it becomes necessary to be able to swap a contract's bytecode for another. Since the new contract might, and probably will, expect a different memory layout than the old contract, a migration procedure for the memory of the contract should also be supported.

This migration procedure is by its nature a destructive operation in terms of the memory of the contract, but may also have consequences for contracts depending on the one being migrated. If the contract changes its externally available API in any way, all contract's that call it should be considered invalid.

#### Possible solution design or implementation
The way to do this might be in the context of a `Session`, introducing a new public function - say `Session::migrate` - taking a contract ID, the new bytecode, and a closure specifying how to migrate from the current memory state to the new state. The arguments to this closure should likely be a handle to instances of both contracts, allowing for several operations to be performed.

```rust
impl Session {
    pub fn migrate<F>(&mut self, contract_id: ContractId, new_bytecode: &[u8], closure: F)
    where
        F:  FnOnce(ContractHandle, ContractHandle) -> Result<()>
    {
         // ...
    }
}
```

ureeves:
> This looks good, how about adding a `dry_run` boolean parameter, so that it is possible to call this method and check if everything is ok? Also, if `dry_run=false`, I'd automatically remove the first contract, in order to prevent the migration from running again.

I understand the spirit of this. Luckily the very concept of a `Session`, together with being able to commit it or not, and us being able to start a new session from before the migration started at any time (rollback), this should not be necessary.

Unless you're talking about checks that should be performed by this function. If so, could you elaborate on what you mean?

miloszm:
Indeed, not committing session solves part of it, but still, we could have custom checks inside of the closure, but I guess there is nothing that session and rollback cannot do, still, it could be more efficient to be able to dry run, and more manageable, esp. if we migrate many contracts, say in a batch. Just sth. to consider.

PR #320, opened 2024-02-05
Add bytecode migrations

ureeves:
Bytecode migrations can now be performed using the `Session::migrate` function. This new function allows for passing in a closure that encodes a migration procedure. The procedure will have both contracts available in the state and is passed the new contract's temporary ID such that it can be called in just the same way as a regular contract. Once the procedure is executed without error the old contract is replaced with the new one and the session returned to the user.

Since it is possible that the `Session` is in an inconsistent state after an error in `Session::migrate`, the function takes the session by value, consuming it, and only returning it if no error was encountered.

Resolves: #313

### DIFF
--- a/contracts/Cargo.toml
+++ b/contracts/Cargo.toml
@@ -6,6 +6,7 @@ members = [
     "constructor",
     "counter",
     "debugger",
+    "double_counter",
     "eventer",
     "everest",
     "fallible_counter",
--- a/contracts/double_counter/Cargo.toml
+++ b/contracts/double_counter/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "double_counter"
+version = "0.1.0"
+authors = [
+    "Eduardo Leegwater Simões <eduardo@dusk.network>",
+]
+edition = "2021"
+
+license = "MPL-2.0"
+
+[dependencies]
+piecrust-uplink = { path = "../../piecrust-uplink", features = ["abi", "dlmalloc"] }
+
+[lib]
+crate-type = ["cdylib", "rlib"]
--- a/contracts/double_counter/src/lib.rs
+++ b/contracts/double_counter/src/lib.rs
@@ -1,0 +1,61 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+//! Contract to implement a simple double counter that can be read and
+//! have either value incremented by one count.
+
+#![no_std]
+
+use piecrust_uplink as uplink;
+
+/// Struct that describes the state of the DoubleCounter contract
+pub struct DoubleCounter {
+    left_value: i64,
+    right_value: i64,
+}
+
+/// State of the DoubleCounter contract
+static mut STATE: DoubleCounter = DoubleCounter {
+    left_value: 0xfc,
+    right_value: 0xcf,
+};
+
+impl DoubleCounter {
+    /// Read the value of the counter
+    pub fn read_values(&self) -> (i64, i64) {
+        (self.left_value, self.right_value)
+    }
+
+    /// Increment the value of the left counter by 1
+    pub fn increment_left(&mut self) {
+        let value = self.left_value + 1;
+        self.left_value = value;
+    }
+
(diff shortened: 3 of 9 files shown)


herr-seppia (approved):
LGTM

However I've some questions:
- should the metadata be replaced too?
- is there a way to prevent a contract to be replaced? Or replaced only in certain circumstances?
- should we support having the closure being a contract itself?

I agree in advance that, except the first, questions can be related to the piecrust consumer (aka rusk)

ureeves:
> LGTM
> 
> However I've some questions:
> 
>     * should the metadata be replaced too?
> 
>     * is there a way to prevent a contract to be replaced? Or replaced only in certain circumstances?
> 
>     * should we support having the closure being a contract itself?
> 
> 
> I agree in advance that, except the first, questions can be related to the piecrust consumer (aka rusk)

I'll answer the questions in order:

- Right now it *can* be replaced. That said we could prevent the function from replacing, if that's what we wish. As if is "should" be or not, I think that's up to downstream to decide, which is essentially how it is right now.
- I'm not sure if I get this one. If you're worried that contracts could call this themselves - making it a clear security nightmare - don't, because they can't.
- This would require an entire different subsystem of imports to be added to support contracts being able to define these transitions. It is possible to do so, but I fail to see a use-case downstream.

These questions seem to be written under the assumption that this functionality is made to be callable by contracts. As I say in the second point, and expand upon in the third, it is most definitely not. The API for the contracts remains exactly the same as before, and only the host can decide to call this under certain circumstances.
My view is that this should be called when we decide to do a genesis contract upgrade at a certain block height. Given that after the procedure the state tree is altered, it is essentially mandatory that all nodes process this upgrade in exactly the same way, meaning that the upgrade code is then essentially a part of the protocol.

herr-seppia:
I've written those questions to figure out if it's possible to handle protocol upgrades without touching the binary.
Indeed, if the clojure could be a contract/wasm, the migration code could be put in the migration-block sidebyside with the new contract code.

But this, indeed, should not be a piecrust problem (we can handle it in rusk properly)

miloszm:
LGTM - A question though - what if we migrate, say, transfer contract, it takes some time, and in the meantime there are some transactions happening in other threads, changing state. Then, when the migration is committed, there could be some data conflicts, or some data could be lost. Should we have some kind of guarantee that other parties are not still using the old contract?

miloszm (approved):

ureeves:
> LGTM - A question though - what if we migrate, say, transfer contract, it takes some time, and in the meantime there are some transactions happening in other threads, changing state. Then, when the migration is committed, there could be some data conflicts, or some data could be lost. Should we have some kind of guarantee that other parties are not still using the old contract?

While it is indeed possible to have multiple `Session`s running in parallel, it is *not possible* to process multiple Transactions in the context of the same session in parallel. This is because every function that writes to a session requires `&mut Session`, and Rust's borrow checker prevents multiple mutable reference from existing. As such the conflicts you speak of are impossible.

On your point about the timing though, you're correct. Such a migration will take time, of the order of the amount of data we want to keep. This will mean that the consensus will need to tolerate the block in which we do this migration taking longer to process than "normal".

miloszm:
> > LGTM - A question though - what if we migrate, say, transfer contract, it takes some time, and in the meantime there are some transactions happening in other threads, changing state. Then, when the migration is committed, there could be some data conflicts, or some data could be lost. Should we have some kind of guarantee that other parties are not still using the old contract?
> 
> While it is indeed possible to have multiple `Session`s running in parallel, it is _not possible_ to process multiple Transactions in the context of the same session in parallel. This is because every function that writes to a session requires `&mut Session`, and Rust's borrow checker prevents multiple mutable reference from existing. As such the conflicts you speak of are impossible.
> 
> On your point about the timing though, you're correct. Such a migration will take time, of the order of the amount of data we want to keep. This will mean that the consensus will need to tolerate the block in which we do this migration taking longer to process than "normal".

That is nice that a single session is safe. How about multiple sessions though? Do we have a possibility of them? - Just asking as I am not sure at the moment. With the timing problem - I think it'd be nice to have some rough estimate how long the migration could take, say, after few months of the blockchain being up with some average traffic.

miloszm:
Could we be in trouble if consensus decides to pick some other state than a given migration was based upon?

ureeves:
> Could we be in trouble if consensus decides to pick some other state than a given migration was based upon?

I mean, then the upgrade is just not performed. I wouldn't say this is "trouble", more like "network choice".

PR #323, opened 2024-02-07
Fix overflow in ICC gas limit calculation

ureeves:


piecrust/src/imports.rs
```diff
@@ -209,7 +209,9 @@ pub(crate) fn c(
     let callee_limit = if gas_limit > 0 && gas_limit < caller_remaining {
```
HDauven, line 209: Small nit unrelated to the actual PR, I don't like that `gas_limit` is on one side here, and in the other compare the comparison operator is flipped.

HDauven (approved):
LGTM

Issue #324, opened 2024-02-08
Switch to bytecheck-secured argument passing in Piecrust uplink helpers


miloszm:
#### Summary
Currently, Piecrust uplink helpers, which are used in contracts for contract methods' argument passing, use unsafe way of deserialising arguments. This is not optimal for the overall integrity and security of the system, as it allows contract callers to maliciously or accidentally send corrupted arguments. In addition, such corruption is undetected and causes harder to track problems.

#### Possible solution design or implementation
We need to use safe code to deserialise contract methods' arguments, and in particular, we should use bytecheck checking in such a way, that an error in bytecheck checking is causing panic.

#### Additional context
This improvement will require changes in Rusk code.

PR #329, opened 2024-02-13
support bytecheck-based integrity check of contract calls' arguments

miloszm:
Support bytecheck-based integrity check of contract calls' arguments,

Implements issue #324

### DIFF
--- a/contracts/counter/src/lib.rs
+++ b/contracts/counter/src/lib.rs
@@ -35,11 +35,11 @@ impl Counter {
 /// Expose `Counter::read_value()` to the host
 #[no_mangle]
 unsafe fn read_value(arg_len: u32) -> u32 {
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -42,7 +42,7 @@ cold-reboot: contracts ## Run the cold reboot test
 
 .PHONY: test contracts cold-reboot assert-counter-contract-small
 
-MAX_COUNTER_CONTRACT_SIZE = 8192
+MAX_COUNTER_CONTRACT_SIZE = 32768
```
ureeves, line 45: This is a **huge** difference. Should we make two different `wrap_call`s available? Say `wrap_call` and `wrap_call_unchecked`?

miloszm: sounds good, I will do that

ureeves (approved):

PR #339, opened 2024-02-22
piecrust: optional owner in `Session::migrate`

### LINKED ISSUE #336: Enhance session's migrate method with optional transfer of ownership
#### Summary
Sessions's method `migrate` allows for convenient migration of data between an old and a new contract, yet typical use case also involves setting the same owner when deploying the new contract. Current implementation requires owner to be explicitly provided, which is not convenient when dealing with, say, genesis contracts, whose ownership is read from the configuration file and in general is not easily accessible at the location where migration is performed.

It would be more optimal if method `migrate`, optionally, provided a possibility to migrate contract ownership as well. This way we could be certain that the new contract has not only the same contract id, but also an owner. It would also be much more convenient to use the method migrate in places where owner is not easily accessible

#### Additional context
This feature is needed, for example, for stake contract migration, due to important optimisations.

ureeves:
`Session::migrate` is changed to allow for taking the owner of the previous contract, as opposed to it being mandatory to set a new owner.

This entailed numerous API change, the gist of which being that the `owner` is handled internally as a `Vec<u8>` as opposed to a `[u8; N]`. This puts some responsibility on the user of the crate to think about what this means for when the contracts call `owner` or `self_owner`. It is likely wise to only set owner which are always the same size.

Resolves #336

### DIFF
--- a/piecrust/CHANGELOG.md
+++ b/piecrust/CHANGELOG.md
@@ -7,8 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `ContractDataBuilder::owner` to allow for setting the owner of a contract
+  on deploy time [#336]
+
 ### Changed
 
+- Make `owner` field optional in `ContractData` and `ContractDataBuilder` [#336]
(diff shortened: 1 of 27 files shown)


miloszm (approved):
LGTM

PR #345, opened 2024-03-18
Allow feeder calls to set the gas limit

### LINKED ISSUE #344: Audit: too much gas for feeder calls
#### Summary

Currently a feeder call is performed with the maximum amount of gas possible. This is ok for contracts that the host controls, but may become a problem for contract it does not. If a contract is malicious it may simply loop forever, and never provide any data, and effectively allow the chain to be DOSed.

https://github.com/dusk-network/piecrust/blob/a6814919dc20347dc1571ec64db04e72056b8e31/piecrust/src/session.rs#L486

#### Possible Solution (Optional)

Allow the caller to determine the gas with which the feeder call can be executed with.

ureeves:
Resolves #344

### DIFF
--- a/piecrust/CHANGELOG.md
+++ b/piecrust/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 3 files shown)


HDauven (approved):
LGTM!

I would like a negative test though, where the gas limit is exceeded.

HDauven (approved):

PR #346, opened 2024-03-18
Fix audit-reported under/overflows

### LINKED ISSUE #343: Audit: possible under/overflows
#### Summary

There are some possible over/underflows which may cause problems when the machine is called under certain conditions:

https://github.com/dusk-network/piecrust/blob/a6814919dc20347dc1571ec64db04e72056b8e31/piecrust/src/imports.rs#L94
https://github.com/dusk-network/piecrust/blob/a6814919dc20347dc1571ec64db04e72056b8e31/piecrust/src/call_tree.rs#L153
https://github.com/dusk-network/piecrust/blob/a6814919dc20347dc1571ec64db04e72056b8e31/piecrust/src/instance.rs#L269

In the spirit of defensive code, the overflows should cause an error, or there should be a very good reason commented as to why this can never happen.

#### Relevant Context (optional)

The first overflow given can be pretty severe if not in a Rust language context, since it could have been used by a contract to go past the boundaries of its own memory and into the memory of the host.

ureeves:
Fixing these required the introduction of an error variant called `Error::ArgumentBufferOverflow`, to handle possible instance of callers calling contracts with way too long arguments.

Resolves #343

### DIFF
--- a/piecrust/src/call_tree.rs
+++ b/piecrust/src/call_tree.rs
@@ -150,6 +150,9 @@ impl Drop for CallTree {
 unsafe fn update_spent(node: *mut CallTreeNode) {
     let node = &mut *node;
     node.children.iter_mut().for_each(|&mut child| unsafe {
+        // It should be impossible for this to underflow since the amount spent
+        // in all child nodes is always less than or equal to the amount spent
+        // in the parent node.
         node.elem.spent -= (*child).elem.spent;
         update_spent(child);
     });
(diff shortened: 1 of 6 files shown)


piecrust/src/call_tree.rs
```diff
@@ -150,6 +150,9 @@ impl Drop for CallTree {
 unsafe fn update_spent(node: *mut CallTreeNode) {
     let node = &mut *node;
     node.children.iter_mut().for_each(|&mut child| unsafe {
+        // It should be impossible for this to underflow since the amount spent
+        // in a child node is always less than the amount spent in the
+        // parent node.
```
moCello, line 155: But is the *sum of all* child.spend also less than the parent.spend?

moCello (approved):

Issue #350, opened 2024-04-04
Economic protocol scenario 3 - contract pays for gas


miloszm:
#### Summary
For economic protocol scenario 3 we need support for contract paying for transaction execution gas rather than the user. In order to achieve that, we need to be able to convey the additional information from the contract to the caller. To such information belongs the decision that contract wants to pay, as well as how much it is willing to pay.

#### Possible solution design or implementation
One possibility is to add an extra buffer B which will convey this information, along with other such meta-information.

#### Additional context
Note that willingness to pay is not something that is constant, it may be a result of a multi-factor decision, hence information about who is willing to pay and how much needs to be carried with every call. Yet, it is a system information, hence it should not be mixed with an "application" information such as a contract method return value.

PR #352, opened 2024-04-10
Changes for Economic Protocol scenario 3 - "contract pays for gas"

miloszm:
Changes for Economic Protocol scenario 3 - "contract pays for gas" and "contract earns fee" parts

Implements issues #350 and #353

### DIFF
--- a/piecrust-uplink/src/abi/state.rs
+++ b/piecrust-uplink/src/abi/state.rs
@@ -4,7 +4,6 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-use alloc::vec::Vec;
 use core::ptr;
 
 use rkyv::{
@@ -15,31 +14,46 @@ use rkyv::{
 };
 
 use crate::{
-    ContractError, ContractId, StandardBufSerializer, CONTRACT_ID_BYTES,
-    SCRATCH_BUF_BYTES,
+    ContractError, ContractId, EconomicMode, RawResult, StandardBufSerializer,
+    CONTRACT_ID_BYTES, ECO_MODE_BUF_LEN, ECO_MODE_LEN, SCRATCH_BUF_BYTES,
 };
 
 pub mod arg_buf {
-    use crate::ARGBUF_LEN;
-
+    use crate::{EconomicMode, ARGBUF_LEN, ECO_MODE_BUF_LEN, ECO_MODE_LEN};
     use core::ptr;
     use core::slice;
 
     #[no_mangle]
     static mut A: [u64; ARGBUF_LEN / 8] = [0; ARGBUF_LEN / 8];
 
+    #[no_mangle]
+    static mut ECO_MODE: [u8; ECO_MODE_BUF_LEN] = [0u8; ECO_MODE_BUF_LEN];
+
     pub fn with_arg_buf<F, R>(f: F) -> R
     where
-        F: FnOnce(&mut [u8]) -> R,
+        F: FnOnce(&mut [u8], &mut [u8]) -> R,
     {
         unsafe {
             let addr = ptr::addr_of_mut!(A);
             let slice = slice::from_raw_parts_mut(addr as _, ARGBUF_LEN);
-            f(slice)
+            let addr_eco_mode = ptr::addr_of_mut!(ECO_MODE);
+            let slice_eco_mode =
+                slice::from_raw_parts_mut(addr_eco_mode as _, ECO_MODE_BUF_LEN);
+            f(slice, slice_eco_mode)
+        }
+    }
+
+    pub fn set_eco_mode(eco_mode: EconomicMode) {
+        unsafe {
+            let addr_eco_mode = ptr::addr_of_mut!(ECO_MODE);
+            let slice_eco_mode =
+                slice::from_raw_parts_mut(addr_eco_mode as _, ECO_MODE_LEN);
+            eco_mode.write(slice_eco_mode);
         }
     }
 }
 
+pub(crate) use arg_buf::set_eco_mode;
 pub(crate) use arg_buf::with_arg_buf;
 
 mod ext {
@@ -73,7 +87,7 @@ where
     Ret: Archive,
     Ret::Archived: Deserialize<Ret, Infallible>,
 {
-    let arg_len = with_arg_buf(|buf| {
+    let arg_len = with_arg_buf(|buf, _| {
         let mut sbuf = [0u8; SCRATCH_BUF_BYTES];
         let scratch = BufferScratch::new(&mut sbuf);
         let ser = BufferSerializer::new(buf);
@@ -88,7 +102,7 @@ where
 
     let ret_len = unsafe { ext::hq(name_ptr, name_len, arg_len) };
 
-    with_arg_buf(|buf| {
+    with_arg_buf(|buf, _| {
         let slice = &buf[..ret_len as usize];
         let ret = unsafe { archived_root::<Ret>(slice) };
         ret.deserialize(&mut Infallible).expect("Infallible")
@@ -132,7 +146,7 @@ where
     Ret: Archive,
     Ret::Archived: Deserialize<Ret, Infallible>,
 {
-    let arg_len = with_arg_buf(|buf| {
+    let arg_len = with_arg_buf(|buf, _| {
         let mut sbuf = [0u8; SCRATCH_BUF_BYTES];
         let scratch = BufferScratch::new(&mut sbuf);
         let ser = BufferSerializer::new(buf);
@@ -155,7 +169,7 @@ where
         )
     };
 
-    with_arg_buf(|buf| {
+    with_arg_buf(|buf, _| {
         if ret_len < 0 {
             Err(ContractError::from_parts(ret_len, buf))
         } else {
@@ -175,8 +189,24 @@ pub fn call_raw(
     contract: ContractId,
     fn_name: &str,
     fn_arg: &[u8],
-) -> Result<Vec<u8>, ContractError> {
+) -> Result<RawResult, ContractError> {
     call_raw_with_limit(contract, fn_name, fn_arg, 0)
+}
+
+/// Allows the contract to set allowance which will be used
+/// to pay for the current call, under the condition that contract's
+/// current balance is greater or equal to the allowance
+/// and the allowance is sufficient to cover gas cost.
+/// This call is of no consequence if the above conditions are not met.
+pub fn set_allowance(allowance: u64) {
+    set_eco_mode(EconomicMode::Allowance(allowance));
+}
+
+/// Allows the contract to set charge which will be used
+/// as a fee. If charge is greater than gas spent,
+/// the difference will be added to contract's balance.
+pub fn set_charge(charge: u64) {
+    set_eco_mode(EconomicMode::Charge(charge));
 }
 
 /// Calls the function with name `fn_name` of the given `contract` using
@@ -192,8 +222,8 @@ pub fn call_raw_with_limit(
     fn_name: &str,
     fn_arg: &[u8],
     gas_limit: u64,
-) -> Result<Vec<u8>, ContractError> {
-    with_arg_buf(|buf| {
+) -> Result<RawResult, ContractError> {
+    with_arg_buf(|buf, _| {
         buf[..fn_arg.len()].copy_from_slice(fn_arg);
     });
 
@@ -210,11 +240,16 @@ pub fn call_raw_with_limit(
         )
     };
 
-    with_arg_buf(|buf| {
+    with_arg_buf(|buf, eco_mode_buf| {
         if ret_len < 0 {
             Err(ContractError::from_parts(ret_len, buf))
         } else {
-            Ok(buf[..ret_len as usize].to_vec())
+            Ok(RawResult::new(
+                buf[..ret_len as usize].to_vec(),
+                EconomicMode::read(
+                    &eco_mode_buf[ECO_MODE_LEN..ECO_MODE_BUF_LEN],
+                ),
+            ))
         }
     })
 }
@@ -234,7 +269,7 @@ where
     unsafe {
         match ext::hd(name, name_len) as usize {
             0 => None,
-            arg_pos => Some(with_arg_buf(|buf| {
+            arg_pos => Some(with_arg_buf(|buf, _| {
                 let ret = archived_root::<D>(&buf[..arg_pos]);
                 ret.deserialize(&mut Infallible).expect("Infallible")
             })),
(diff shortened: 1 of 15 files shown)


piecrust-uplink/src/abi/state.rs
```diff
@@ -19,15 +19,24 @@ use crate::{
     SCRATCH_BUF_BYTES,
 };
 
+pub enum BIndex {
+    Allowance = 0,
+    Charge = 1,
+}
```
ureeves, line 25: Maybe we could avoid making this public? Or we could make a richer type, for example:

```rust
/// Probably choose a different name for the enum 😅
pub enum EconomicChoice {
    // Different regimes that can be chosen by the contract.
    Allowance(u64),
    Charge(u64),
    None
}
```

piecrust/src/session.rs
```diff
@@ -777,8 +778,20 @@ impl Session {
...
+        let mut spent = limit - instance.get_remaining_gas();
+
+        if allowance > 0 {
+            // the call has been paid for by the contract so we can set
+            // the spent amount to zero,
+            spent = 0;
+        } else {
+            // add possible contract's charge
+            spent += charge;
+        }
```
ureeves, line 794: Wouldn't it make more sense to modify the `CallReceipt` to include what the contract set in the economic buffer? This instead of modifying the gas spent.

ureeves (changes requested):
I've got some things I''d like to see some work on, and also some broader remarks.

My main concern this mechanism's implementation is that the contract must remember to **always** set the value of the economic buffer at the end of a function it exports. This because otherwise the previous value will be used. This is and of itself is not a problem, but it does become an issue when we don't make this clear/ergonomic to the user.

We can choose one of two options afaics:

- Have the host set the "default" value every time the contract is called, erasing the need for the contract to do it.
- Modify the `wrap_call` functions to make this ergonomic to the user.

There's other things interspersed below, but one thing I'd like to mention is an aesthetic concern - what do we name what is now called "buffer b"? Certainly we don't want to just keep that name. Maybe "economic buffer"?

miloszm:
Re: _My main concern this mechanism's implementation is that the contract must remember to **always** set the value of the economic buffer at the end of a function it exports. This because otherwise the previous value will be used. This is and of itself is not a problem, but it does become an issue when we don't make this clear/ergonomic to the user._

Problem of previous value is already taken care of, economic buffer is set to zero before every call, previous value is not used, contract has to set allowance or charge only if he chooses to do so, otherwise does it not have to do anything. See session.rs line 765

As for other remarks, they are good remarks IMO and I will follow on most of them. Not sure about CallReceipt remark yet, will need to analyse it more and provide feedback after that.

piecrust-uplink/src/abi/state.rs
```diff
@@ -179,6 +210,33 @@ pub fn call_raw(
...
+/// Same as [`call_raw`] but in addition to a called
+/// function return value, it also returns the contents
+/// of buffer B
+pub fn call_raw_b(
+    contract: ContractId,
+    fn_name: &str,
+    fn_arg: &[u8],
+) -> Result<(Vec<u8>, u64, u64), ContractError> {
+    call_raw_with_limit_b(contract, fn_name, fn_arg, 0)
+}
```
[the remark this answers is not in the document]
miloszm: done

piecrust-uplink/src/abi/state.rs
miloszm: done

piecrust-uplink/src/abi/state.rs
```diff
@@ -219,6 +277,45 @@ pub fn call_raw_with_limit(
...
 
+/// Same as [`call_raw_with_limit`] but in addition to a called
+/// function return value, it also returns the contents
+/// of buffer B
+pub fn call_raw_with_limit_b(
+    contract: ContractId,
+    fn_name: &str,
+    fn_arg: &[u8],
+    gas_limit: u64,
+) -> Result<(Vec<u8>, u64, u64), ContractError> {
```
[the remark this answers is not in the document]
miloszm: done

piecrust/src/imports.rs
```diff
@@ -246,15 +248,18 @@ pub(crate) fn c(
         let name = core::str::from_utf8(&memory[name_ofs..][..name_len])?;
 
         let arg = &arg_buf[..arg_len as usize];
+        let arg_b = &arg_buf_b[..ARGBUF_B_LEN];
 
         callee.write_argument(arg);
+        callee.write_argument_b(arg_b);
```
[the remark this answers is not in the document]
miloszm: done

piecrust-uplink/src/abi/state.rs
```diff
@@ -179,6 +210,33 @@ pub fn call_raw(
...
+pub fn set_allowance(b: u64) {
+    set_b(b, BIndex::Allowance);
+}
+
+/// Allows the contract to set charge which will be used
+/// as a fee. If charge is greater than gas spent,
+/// the difference will be added to contract's balance.
+pub fn set_charge(b: u64) {
+    set_b(b, BIndex::Charge);
+}
```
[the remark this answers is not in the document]
miloszm: I decided to leave these functions as separate ones, because I think they are more contract writer-friendly this way IMO. Let's give it a second look.

piecrust/src/session.rs
miloszm: I evaluated this suggestion and I came to the conclusion that it is important to set the `spent` variable at this particular place. IMO data like allowance or charge cannot go to CallReceipt because this functionality is not optional. System needs to act as if gas was either not consumed (allowance) or consumed with additional charge, so that from the point of view of the caller the behaviour is completely deterministic. If we leave the allowance/charge in `CallReceipt`, we lose control about enforcement of our policy. We cannot be sure if it will be enforced, yet, since we charged/deposited the contract with actual money already at this point, we must make the enforcement final at this point as well.

ureeves (changes requested):
I think we should introduce an easy way for the contract to set the economic mode after each exposed function. My proposal would be to have it be done using the `wrap_call`  helpers that we already have.

This is because the economic buffer is copied from any contract called, and then passed along. To make sure that no contract makes a decision on how to spend the funds for another, I think this should definitely be provided.

piecrust/src/session.rs
ureeves: `spent` contains information about gas spent, and any additional data would contain information the economic buffer of a called contract. We don't lose control at all on enforcing any policy currently existing - not sure what you're referring to with this - and allow for more freedom downstream on how to handle it.

piecrust-uplink/src/abi/state.rs
```diff
@@ -15,31 +14,43 @@ use rkyv::{
...
+            let addr_eco_args = ptr::addr_of_mut!(ECO_ARGS);
+            let slice_eco_args =
+                slice::from_raw_parts_mut(addr_eco_args as _, ECO_ARGS_LEN / 8);
+            f(slice, slice_eco_args)
+        }
+    }
+
+    pub fn set_eco_arg(arg: u64, index: EconomicMode) {
+        unsafe {
+            ECO_ARGS[index as usize] = arg;
```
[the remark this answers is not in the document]
miloszm: charge and allowance are not possible at the same time, I made the change according to your suggestions

piecrust/src/session.rs
miloszm: I do not see any freedom about how to handle it downstream, as the money is already moved at this point, so all we need to do is to even it out, otherwise we may cause economic imbalance - but let's discuss it

miloszm: done

piecrust/src/imports.rs
```diff
@@ -248,13 +249,20 @@ pub(crate) fn c(
...
         check_arg(callee, ret_len as u32)?;
 
         // copy back result
         callee.read_argument(&mut memory[argbuf_ofs..][..ret_len as usize]);
+        callee.read_eco_mode_buf(&mut memory[eco_mode_ofs..][..ECO_MODE_LEN]);
+        // if depth of the stack is 3 or more, economic mode is not read
+        const ECO_STACK_DEPTH: usize = 2;
+        if env.nth_from_top(ECO_STACK_DEPTH).is_some() {
+            callee.clear_eco_mode();
+        }
```
ureeves, line 265: I really don't think this should be the mechanism. While I don't think contracts should be able to make decisions for others, I also don't think we should stop contracts from reading what a callee puts in their buffer.

This feels like a bandaid for a deeper issue. I suggest either having a double buffer - one for writing by the contract and reading by the host, and another for reading by the contract and writing by the host - or make it easy for contracts to *always* specify the economic mode - by modifying the `wrap_call` functions.

ureeves (changes requested):

piecrust-uplink/src/types.rs
```diff
@@ -134,3 +136,83 @@ impl core::fmt::Display for ContractId {
...
+    Charge(u64),
+    None,
+    Unknown,
+}
+
+impl EconomicMode {
+    const ALLOWANCE: u8 = 1;
+    const CHARGE: u8 = 2;
+    const NONE: u8 = 0;
+    const UNKNOWN: u8 = 3;
```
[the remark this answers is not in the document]
miloszm: That is exactly my opinion as well

piecrust-uplink/src/abi/state.rs
```diff
@@ -15,31 +14,46 @@ use rkyv::{
...
+            f(slice, slice_eco_mode)
+        }
+    }
+
+    pub fn set_eco_mode(eco_mode: EconomicMode) {
+        unsafe {
+            let addr_eco_mode = ptr::addr_of_mut!(ECO_MODE);
+            let slice_eco_mode =
+                slice::from_raw_parts_mut(addr_eco_mode as _, ECO_MODE_LEN);
+            eco_mode.write(slice_eco_mode);
```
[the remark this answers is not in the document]
miloszm: I think we should not have `unknown` at all

piecrust/src/imports.rs
ureeves: I strongly disagree with this characterization of the problem. The issue exists whether we consider downstream or not. The bottom line is that a callee should not be able to set the callers economic mode, regardless of call stack height. As it stands this is exactly what happens, but we make an exception for call stack heights larger than 2, which is why I call this a bandaid.

A double buffer would totally solve the problem. You could even just double the size of the buffer and treat each half differently - which would be my favorite approach. Expanding:

```text
                     1                   2                   3
 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2

Caller
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|          Write Buffer          |           Read Buffer          |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
                                                  |
                +------->-------->-------->--------
Callee          |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|          Write Buffer          |           Read Buffer          |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```

The contract writes to the write buffer, and reads from the read buffer, while the host reads from the write buffer, and writes to the read buffer. This means that any callee will effectively only ever set the read buffer of the caller, and as such the caller's decision can never be influence by the callee. This, together with zeroing the contract's write buffer - ensuring the previous decisions don't influence the current one - would serve perfectly well as the economic mode mechanism.

I also think it is a bad design decision to couple `piecrust` with downstream's chosen protocol call stack height. We might as well merge this repository with `rusk` at that point.

piecrust-uplink/src/types.rs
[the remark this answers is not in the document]
ureeves: I think there's a misunderstanding here. I think the enum variant should still exist, I just think the tag is meaningless, since we should never really *write* it. Only when *reading*, when we don't understand what is on there, should we instantiate `EconomicMode::Unknown`, allowing the caller to then branch based on it not understanding what is in the buffer.

piecrust/src/imports.rs
miloszm: I am also very interested in keeping Piecrust architecturally sound and agnostic to such peculiarities, yet I need a scheme that works. As far as I can see, it is not possible to come up with a scheme which will not take call stack height into account (or some kind of exception for height 1). This is a derivative of having transfer contract (as a contract). Other solution would be to move the EP logic into host, but I think this would be even more inconsistent with our overall design.

miloszm: Also, I do not think this is as bad as being equivalent of merging `Piecrust` into `rusk`. We need a quite general parameter to Piecrust which would mean: `number of layers of privileged contracts`. In our case it would be `1`, as we have one layer of privileged contracts, i.e., a transfer contract, which is calling all other contracts. And Piecrust would have some buffers accessible only to privileged contracts, that is basically the idea.

piecrust-uplink/src/types.rs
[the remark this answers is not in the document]
miloszm: you wrote: _I don't think it makes much sense to have a tag for EconomicMode::Unknown_, but you really meant `const UNKNOWN: u8 = 3;` right? So you are for keeping the EconomicMode::Unknown and for removing the const right?

piecrust-uplink/src/abi/state.rs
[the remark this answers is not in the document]
miloszm: ok, I will keep EconomicMode::Unknown and panic when it is about to be written

piecrust/src/imports.rs
ureeves: No, the scheme does not collapse in this case. The transfer contract is always free to - and likely will - write what it read from the "read buffer" to its own "write buffer" so it can be picked up by the host if necessary. It simply becomes a positive, conscious (i.e. by its code), decision, instead of having been perhaps inadvertently set by a contract that it happened to call. 

BTW in the diagram the callee is at the bottom, not at the top, and as such the transfer contract would be the top contract. Reference frame problem :smile: 

While I agree with you that keeping this is not as bad as merging the whole repository into `rusk`, it is a very significant coupling that I think we should avoid. To understand how significant it is, one just has to think of how (if it was merged) the proxy contract PR would have had an influenced `piecrust`, had this already been merged.

miloszm: I see that the difference between our approaches is that I think in terms of what is possible (or allowed), and you think in terms of what can be consciously done (or not easily). If the intention is to make it less accidental but possible, then your scheme will work. I wanted to have a scheme which does not allow callees beyond 2nd level to set the EP arguments at all. Yet if we assume they can do it and the only assumption is that the info is consciously read by callers - the scheme you proposed is doable. Glad this is settled - I will implement it and let you know.

ureeves (approved):
LGTM!

Issue #351, opened 2024-04-08
Migrate from Cranelift to Winch


HDauven:
#### Summary
With Cranelift, it is currently possible for a malicious actor to do a JIT bomb. Wasmtime currently provides a single-pass compiler that should prevent this called Winch.

#### Possible solution design or implementation
Switch from Cranelift to Winch to prevent potential JIT bombs.

ureeves:
This cannot be implemented until `wasmtime` implements fuel-based interruption
https://github.com/bytecodealliance/wasmtime/issues/8090

PR #354, opened 2024-04-15
Upgrade to `dusk-wasmtime` version `19`

ureeves:
In this version, `wasmtime` introduces a preliminary version of the [`winch`] compiler, which is ready to use on `x86` embeddings. They advise not using it in production just yet, but at least its ready for testing.

[`winch`]: https://github.com/bytecodealliance/wasmtime/releases/tag/v19.0.0

See-also: https://github.com/dusk-network/piecrust/issues/351


herr-seppia (approved):
LGTM

Issue #359, opened 2024-05-10
Pricing strategy for Host Functions and Storage


HDauven:
#### Summary

Currently, host functions do not have a "fair" price attached to them. Storage is priced solely based on the computation of memory allocation.

This approach is most likely not optimal as it does not account for the actual resource usage and the economic impact of host functions computational load and underpriced storage.

#### Possible solution design or implementation

To address this, we should figure out the computational complexity of host functions (Accounting for network load and implicit storage as well?) and figure out an optimum of storage versus computational cost. Ideas like expiration and rent can be considered, but would be out of scope for now.

PR #361, opened 2024-05-15
Make storage instructions cost

ureeves:
Each byte stored in the WebAssembly linear memory of a contract is made to cost 4 units of fuel - in the units that wasmtime uses. This makes storing a single byte as expensive as, for example, multiplying or subtracting two numbers.

The `MemoryGrow` instruction is purposefully left with the default price - just like any other instruction, due the fact that expanding the size of the linear memory doesn't actually cost a lot of resources, owing to the fact that a page that is never written to is never written to disk.

A single contract's compiled bytecode will contain a header, encoding within it information about the cost per WebAssembly instruction. If these don't match the costs configured for the current `VM` instantiation, the contract will be recompiled, just as what happens if the native instruction set doesn't match.

See-also: https://github.com/dusk-network/piecrust/issues/359

### DIFF
--- a/piecrust/CHANGELOG.md
+++ b/piecrust/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Changed
 
+- Make storage instructions cost 4 gas per byte [#359]
 - Upgrade `dusk-wasmtime` to version `21.0.0-alpha` [#359]
 
 ## [0.19.0] - 2024-05-08
--- a/piecrust/CHANGELOG.md
+++ b/piecrust/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


HDauven (approved):
Awesome, LGTM!

herr-seppia:
> A single contract's compiled bytecode will contain a header, encoding within it information about the cost per WebAssembly instruction

Is this part of the PR? I haven't found any change for this

ureeves:
> > A single contract's compiled bytecode will contain a header, encoding within it information about the cost per WebAssembly instruction
> 
> Is this part of the PR? I haven't found any change for this

No its not. It was a part of this one https://github.com/dusk-network/piecrust/pull/288. I thought it deserved a mention though, to make it clear.

ureeves:
> > A single contract's compiled bytecode will contain a header, encoding within it information about the cost per WebAssembly instruction
> 
> Is this part of the PR? I haven't found any change for this

No, it was a part of [this one](https://github.com/dusk-network/wasmtime/pull/1), and the reason why we had an upgrade of the dusk-wasmtime version during this PR.

PR #362, opened 2024-05-16
Allow specifying host query price

ureeves:
Host queries are modified to be able to specify their price. The calling of a such a query proceeds in two stages - deserializing and pricing, and then execution. In between the two stages, `piecrust` is able to make the choice of whether to continue with execution or not, based on the gas remaining for the execution.

To achieve this, the `HostQuery` trait is modified by adding two functions - `deserialize_and_price` and `execute` - and by removing the `Fn(&mut [u8], u32) -> u32` bound. This structure allows the implementer to price queries fairly, as well as prevent double deserialization by leveraging `Box<dyn Any>` to pass data between the two functions.

`HostQuery` remains implemented for all `Fn(&mut [u8], u32) -> u32` that are also `Send` and `Sync`, so this is not a breaking change.

See-also: https://github.com/dusk-network/piecrust/issues/359

### DIFF
--- a/piecrust/CHANGELOG.md
+++ b/piecrust/CHANGELOG.md
@@ -7,8 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Added
+
+- Apply charging mechanism for host queries [#359]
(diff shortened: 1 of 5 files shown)


HDauven (approved):
LGTM!

fed-franz (commented):
LGTM, but I'm not confident with the code, so I'll just "Comment" :)

Issue #357, opened 2024-05-01
Enrich deploy method and contract's metadata with optional gas limit and gas price hint for free transactions


miloszm:
#### Summary
In order to be able to execute deposit-free free transactions, we need to assume some gas limit and some gas price hint, which are normally, in paid transactions, given by the user. Block generator needs to be able to obtain gas limit and to calculate gas price from a price hint (a ratio to the average gas price).

#### Possible solution design or implementation
Deploy method needs to accept optional arguments for gas limit and gas price hint and store them in contract's metadata, along with contract's owner.

#### Additional context
This is part of Economic Protocol scenario 3 implementation, related to issues #350 and #353.

PR #363, opened 2024-05-29
Enrich deploy method and contract's metadata with optional gas limit and gas price hint for free transactions

miloszm:
Enrich deploy method and contract's metadata with optional gas limit and gas price hint for free transactions,

Implements issue #357

### DIFF
--- a/piecrust-uplink/src/abi/state.rs
+++ b/piecrust-uplink/src/abi/state.rs
@@ -301,7 +301,7 @@ pub fn owner<const N: usize>(contract: ContractId) -> Option<[u8; N]> {
 pub fn free_limit(contract: ContractId) -> Option<u64> {
     let contract_id_ptr = contract.as_bytes().as_ptr();
     unsafe {
-        match ext::free_limit(contract_id_ptr) as usize {
+        match ext::free_limit(contract_id_ptr) {
             0 => None,
             _ => with_arg_buf(|buf, _| {
                 if buf[0] == 0 {
@@ -322,7 +322,7 @@ pub fn free_price_hint(contract: ContractId) -> Option<(u64, u64)> {
     let contract_id_ptr = contract.as_bytes().as_ptr();
 
     unsafe {
-        match ext::free_price_hint(contract_id_ptr) as usize {
+        match ext::free_price_hint(contract_id_ptr) {
             0 => None,
             _ => with_arg_buf(|buf, _| {
                 if buf[0] == 0 {
--- a/piecrust-uplink/src/abi/state.rs
+++ b/piecrust-uplink/src/abi/state.rs
@@ -18,6 +18,8 @@ use crate::{
     CONTRACT_ID_BYTES, ECO_MODE_BUF_LEN, ECO_MODE_LEN, SCRATCH_BUF_BYTES,
 };
 
+const LEN_U64: usize = core::mem::size_of::<u64>();
+
 pub mod arg_buf {
     use crate::{EconomicMode, ARGBUF_LEN, ECO_MODE_BUF_LEN, ECO_MODE_LEN};
     use core::ptr;
@@ -301,9 +303,14 @@ pub fn free_limit(contract: ContractId) -> Option<u64> {
     unsafe {
         match ext::free_limit(contract_id_ptr) as usize {
             0 => None,
-            arg_pos => with_arg_buf(|buf, _| {
-                let ret = archived_root::<Option<u64>>(&buf[..arg_pos]);
-                ret.deserialize(&mut Infallible).expect("Infallible")
+            _ => with_arg_buf(|buf, _| {
+                if buf[0] == 0 {
+                    None
+                } else {
+                    let mut value_bytes = [0; LEN_U64];
+                    value_bytes.copy_from_slice(&buf[1..LEN_U64 + 1]);
+                    Some(u64::from_le_bytes(value_bytes))
+                }
             }),
         }
     }
@@ -317,9 +324,18 @@ pub fn free_price_hint(contract: ContractId) -> Option<(u64, u64)> {
     unsafe {
         match ext::free_price_hint(contract_id_ptr) as usize {
             0 => None,
-            arg_pos => with_arg_buf(|buf, _| {
-                let ret = archived_root::<Option<(u64, u64)>>(&buf[..arg_pos]);
-                ret.deserialize(&mut Infallible).expect("Infallible")
+            _ => with_arg_buf(|buf, _| {
+                if buf[0] == 0 {
+                    None
+                } else {
+                    let mut value_bytes = [0; LEN_U64];
+                    value_bytes.copy_from_slice(&buf[1..LEN_U64 + 1]);
+                    let num = u64::from_le_bytes(value_bytes);
+                    value_bytes
+                        .copy_from_slice(&buf[LEN_U64 + 1..LEN_U64 * 2 + 1]);
+                    let denom = u64::from_le_bytes(value_bytes);
+                    Some((num, denom))
+                }
             }),
         }
     }
--- a/piecrust-uplink/src/abi/state.rs
+++ b/piecrust-uplink/src/abi/state.rs
@@ -76,6 +76,8 @@ mod ext {
         pub fn limit() -> u64;
         pub fn spent() -> u64;
         pub fn owner(contract_id: *const u8) -> i32;
+        pub fn free_limit(contract_id: *const u8) -> i32;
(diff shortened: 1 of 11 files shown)


piecrust-uplink/src/abi/state.rs
```diff
@@ -292,6 +294,37 @@ pub fn owner<const N: usize>(contract: ContractId) -> Option<[u8; N]> {
...
+/// Returns given contract's free limit, if the contract exists and if it
+/// has a free limit set.
+pub fn free_limit(contract: ContractId) -> Option<u64> {
+    let contract_id_ptr = contract.as_bytes().as_ptr();
+    unsafe {
+        match ext::free_limit(contract_id_ptr) as usize {
+            0 => None,
+            arg_pos => with_arg_buf(|buf, _| {
+                let ret = archived_root::<Option<u64>>(&buf[..arg_pos]);
+                ret.deserialize(&mut Infallible).expect("Infallible")
```
ureeves, line 306: I would like to prevent the addition of `rkyv` into the ABI if possible. This is due to `rkyv` being fairly Rust-specific. I would suggest coming up with a custom-yet-sane™ serialization for this integer. Can we use the return of the `extern` instead?

piecrust-uplink/src/abi/state.rs
```diff
@@ -292,6 +294,37 @@ pub fn owner<const N: usize>(contract: ContractId) -> Option<[u8; N]> {
...
+/// if it has a free price hint set.
+pub fn free_price_hint(contract: ContractId) -> Option<(u64, u64)> {
+    let contract_id_ptr = contract.as_bytes().as_ptr();
+
+    unsafe {
+        match ext::free_price_hint(contract_id_ptr) as usize {
+            0 => None,
+            arg_pos => with_arg_buf(|buf, _| {
+                let ret = archived_root::<Option<(u64, u64)>>(&buf[..arg_pos]);
+                ret.deserialize(&mut Infallible).expect("Infallible")
```
ureeves, line 322: The same applies here, but I can see it being more complicated due to it being a tuple.

piecrust/src/imports.rs
```diff
@@ -452,15 +457,24 @@ fn owner(fenv: Caller<Env>, mod_id_ofs: usize) -> WasmtimeResult<i32> {
             mod_id
         });
 
+        if env.instance(&mod_id).is_none() {
+            let _ = env.create_instance(mod_id);
+        }
```
ureeves, line 462: I can't understand this construction. Could you clarify?

ureeves (changes requested):
`rkyv` should be avoided in the host-contract ABI, due to it being too Rust-specific

piecrust/src/imports.rs
miloszm: Without this construction getting metadata (even including owner) fails before the first call to a contract. For price hint metadata, which we need to know before calling the contract, it is critical to have it available all the time, not only after first call. In general, I think such behaviour is a bug. If you can suggest a different fix for this, I'd be happy to see it.

piecrust-uplink/src/abi/state.rs
miloszm: sounds good, I'll do it

piecrust-uplink/src/abi/state.rs
miloszm: sounds good

piecrust-uplink/src/abi/state.rs
miloszm: done

piecrust-uplink/src/abi/state.rs
miloszm: done

piecrust/src/imports.rs
ureeves: Oh, I understand now. Ideally we wouldn't need to instantiate it to get at the contract. The metadata should be available in the store, not the instance.

piecrust-uplink/src/abi/state.rs
```diff
@@ -292,6 +296,51 @@ pub fn owner<const N: usize>(contract: ContractId) -> Option<[u8; N]> {
...
+                    None
+                } else {
+                    let mut value_bytes = [0; LEN_U64];
+                    value_bytes.copy_from_slice(&buf[1..LEN_U64 + 1]);
+                    Some(u64::from_le_bytes(value_bytes))
+                }
+            }),
+        }
+    }
+}
```
ureeves, line 317: Do we need to use the first byte to signal the `None`? It seems we can use only the return of the function.

```suggestion
/// Returns given contract's free limit, if the contract exists and if it
/// has a free limit set.
pub fn free_limit(contract: ContractId) -> Option<u64> {
    let contract_id_ptr = contract.as_bytes().as_ptr();
    unsafe {
        match ext::free_limit(contract_id_ptr) {
            0 => None,
            _ => with_arg_buf(|buf, _| {
                let mut value_bytes = [0; LEN_U64];
                value_bytes.copy_from_slice(&buf[0..LEN_U64]);
                Some(u64::from_le_bytes(value_bytes))
            }),
        }
    }
}
```

piecrust-uplink/src/abi/state.rs
```diff
@@ -292,6 +296,51 @@ pub fn owner<const N: usize>(contract: ContractId) -> Option<[u8; N]> {
...
+                    let num = u64::from_le_bytes(value_bytes);
+                    value_bytes
+                        .copy_from_slice(&buf[LEN_U64 + 1..LEN_U64 * 2 + 1]);
+                    let denom = u64::from_le_bytes(value_bytes);
+                    Some((num, denom))
+                }
+            }),
+        }
+    }
+}
```
ureeves, line 342: Exact same thing here:

```suggestion
/// Returns given contract's free gas price hint, if the contract exists and
/// if it has a free price hint set.
pub fn free_price_hint(contract: ContractId) -> Option<(u64, u64)> {
    let contract_id_ptr = contract.as_bytes().as_ptr();

    unsafe {
        match ext::free_price_hint(contract_id_ptr) {
            0 => None,
            _ => with_arg_buf(|buf, _| {
                let mut value_bytes = [0; LEN_U64];
                value_bytes.copy_from_slice(&buf[0..LEN_U64]);
                let num = u64::from_le_bytes(value_bytes);
                value_bytes
                    .copy_from_slice(&buf[LEN_U64..LEN_U64 * 2]);
                let denom = u64::from_le_bytes(value_bytes);
                Some((num, denom))
            }),
        }
    }
}
```

ureeves (changes requested):
The ABI of `ext::free_limit` and `ext::free_price_hint` can solely be used to signal a `None` as opposed to also using the first byte in the argument buffer. It feels like repetition to use both. Are you trying to include information about the existance of the contract in the ABI?

Also, the cast to `usize` is unnecessary I think.

piecrust-uplink/src/abi/state.rs
miloszm: yes, we convey existence of contract and also existence of the setting, hence, we need to encode an option separately, so the above suggestion is not good IMO

piecrust-uplink/src/abi/state.rs
miloszm: same here, we have to pass option if the contract exists, so this is a double option

piecrust/src/imports.rs
miloszm: Problem is that when calling `contract_metadata` we don't know if contract has been loaded, i.e., it exists in the commit, but its metadata might still be missing in ContractSession contracts. One solution would be to do something like:
```
pub fn contract_metadata(
        &mut self,
        contract_id: &ContractId,
    ) -> Option<&ContractMetadata> {
        let _ = self.contract(contract_id.clone());
        self.contracts
            .get(contract_id)
            .map(|store_data| store_data.metadata.data())
    }
```
Yet it has an ugly side that it works on `&mut self` now. Another solution would be to preload all contracts up front, but that could be very wasteful. So I tend to think that the above solution is lesser evil. Basically, we have a caching situation here, hence, an apparently "read-only" function is "read/write". Or we could call env.memory_len(mod_id) instead as a quick solution, although cryptic and surprising for future reader (we could rename memory_len to sth more revealing). What do you think?

miloszm: So in line 461 we could have:
`let _ = env.memory_len(mod_id);`
or, if we copy and rename this method:
`env.load_metadata(mod_id);`

miloszm: After an off-line conversation we decided to go with the above solution.

ureeves (approved):
LGTM

Issue #365, opened 2024-06-10
Removal of the *charge* portion of the economic model 


ureeves:
#### Summary

In the streamlining of the economic protocol described in https://github.com/dusk-network/rusk/issues/1644, we've decided to do away with the charging portion, and as such we need to remove such functionality here as well.

#### Detailed Description (optional)

By merging #363 we introduced a charging component, recognized by the ABI, and formalized into an enum:

https://github.com/dusk-network/piecrust/blob/8502e9e1ce0a689a04e7992b12e36a6b0c27176c/piecrust-uplink/src/types.rs#L162-L169  

We would like to remove the `Charge` variant, and any code relevant to it. This will mean slight changes in the ABI, particularly in which modes are recognized. The `set_charge` function should also be removed.

PR #366, opened 2024-06-11
Removal of the 'charge' portion of the Economic Model scenario 3

miloszm:
Removal of the 'charge' portion of the Economic Model scenario 3

Implements issue #365


ureeves (approved):
LGTM

PR #370, opened 2024-07-01
Price event bytes at the same rate as storage

### LINKED ISSUE #359: Pricing strategy for Host Functions and Storage

ureeves:
This PR introduces a cost to event emission proportional to the amount of bytes emitted, and at the same rate as a storage instruction. The rationale is that, while nodes are not necessarily required by protocol to hold this data like normal storage, it is still crucial data to hold for external applications, and contracts should be discouraged from emitting events that are "too large".

Hence, a payment at the same rate of storage seems appropriate for now. We may always choose to fine-tune this in a future protocol upgrade.

This should be the final round of pricing PRs, as I can't see the sense in pricing any other WebAssembly import offered to the contracts.

Resolves #359

### DIFF
--- a/contracts/eventer/src/lib.rs
+++ b/contracts/eventer/src/lib.rs
@@ -8,6 +8,9 @@
 
 #![no_std]
 
+extern crate alloc;
+use alloc::vec::Vec;
+
 use piecrust_uplink as uplink;
 
 /// Struct that describes the state of the eventer contract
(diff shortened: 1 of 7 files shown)


ureeves:
> LGTM
> 
> Just to clarify, events are supposed to be stored in RocksDB for archival nodes and not as part of the smart contract storage, correct?
> 
> If this is the case, I believe we should consider incentivizing the use of events over storing data directly on the contract by making the price lower from the start.
> 
> Encouraging the use of events early on would help with reducing state bloat due to it being a cheaper option for certain use cases.

The lowest we're getting event storage is not pricing the `emit` import at all and just relying on the fact that the contract has to write to the argument buffer - using storage instructions.

To be honest, if I look at all of this, prices are pretty low regardless of the choice. It's just the relative value that's important.

fed-franz (approved):
LGTM

Issue #375, opened 2024-07-26
Add `impl PartialEq<[u8; 32]> for ContractId`


ureeves:
#### Summary

Implementing the `PartialEq` trait, and its permutations, for `[u8; 32]` and `ContractId` will allow for downstream comparisons that now have to go through the `ContractId::to_bytes` or `ContractId::from_bytes` functions to be much more concise and easy to read. 

#### Possible Solution (optional)

```rust
impl PartialEq<[u8; 32]> for ContractId
impl PartialEq<ContractId> for [u8; 32] // if possible
// to consider
impl<'a, 'b> PartialEq<&'b [u8; 32]> for &'a ContractId
impl<'a> PartialEq<[u8; 32]> for &'a ContractId
// ...
```

PR #376, opened 2024-07-29
uplink: add `impl PartialEq<[u8; 32]> for ContractId`

ureeves:
According to the [cross crate considerations](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html#cross-crate-considerations) of the Rust documentation, one should only implement `PartialEq` for local types, and not for foreign types.

As for implementing it on references... `PartialEq` already takes both operands by reference, making these implementations useless in practice.

Resolves: #375


HDauven (approved):
LGTM

PR #380, opened 2024-08-21
Fix behavior on calling non-existent contracts

ureeves:
The path when a contract calls another contract which does not exist was not handled, and would lead to a panic of the entire execution when it does. This PR fixes this by introducing a new error path in the `c` import that *does not* revert, and a new `ContractError::DoesNotExist` variant.

### DIFF
--- a/contracts/double_counter/src/lib.rs
+++ b/contracts/double_counter/src/lib.rs
@@ -10,6 +10,7 @@
 #![no_std]
 
 use piecrust_uplink as uplink;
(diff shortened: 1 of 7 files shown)


miloszm (approved):
Tested it, it fixes the issue

HDauven (approved):
Very interesting. I seem to recall there being handling for this but it's good that we have a test now. 👍

If we can come up with more edge cases to test, even better. :smile:

PR #381, opened 2024-08-21
Change constructor to init(ializer)

Neotamandua:
The reasons for this change are mainly to improve consistency and also to correctly refer to the init function in the ContractDataBuilder.

Also, my understanding is that the init function is a generic function that is run once during deployment to execute any initialization logic. This implies to me that it can either execute init logic, including initializing data within a contract, but it does not construct the contract, an object, or the state (which already exists after compilation through the static mut STATE in most contracts).

I have also changed the naming within the example contracts. I initially suggested changing constructor_arg to initializer_arg, but opted for init_arg to avoid the z & s ambiguity in UK vs US English.


contracts/Cargo.toml
```diff
@@ -3,11 +3,11 @@ members = [
     "box",
     "c-example",
     "callcenter",
-    "constructor",
+    "initializer",
```
HDauven, line 6: It is pretty amusing you mention the 'z', 's' ambiguity but end up having to use it here, haha.

HDauven (approved):
LGTM, but I would wait for @ureeves or @miloszm to also give their blessing.

ureeves (approved):
Always good to disambiguate! LGTM

contracts/Cargo.toml
ureeves: Hey, he likes it the American way as opposed to the British one. Can't say I blame him. 🤪

PR #382, opened 2024-09-04
Add additional "raw" calls

ureeves:
This PR adds two other "raw" calls in `piecrust-uplink`, meant to allow the user to perform their own custom serialization in contract. `emit_raw` and `feed_raw` make sure that any serialization that the contract developer wishes to use is supported, as opposed to only `rkyv`, in the same way as `call` and `call_raw` already work.


herr-seppia (approved):
LGTM

PR #383, opened 2024-09-04
Remove `'static` bound on `emit` topics

ureeves:
Pretty self-explanatory. The `'static` bound was there because we though users should be stimulated to define their event types, but since this is not a real constraint, we remove it.


Neotamandua (approved):
LGTM

Issue #342, opened 2024-03-11
Make comitting `O(N)`, as opposed to `O(S)`


ureeves:
#### Summary

Currently, in an effort to make commits independent from others, we hard link each page that doesn't change from the previous(base) commit from every contract. This means that during each commit a number of operation is performed that is of the total size of the state (`O(S)`).

This is clearly undesirable, since having the time to commit scale linearly with the size of the overall state means that it will take a long time to commit for larger states. Not only that but over time, while the state grows, we may find ourselves exceeding the threshold of performance allowed by the consensus, simply for having the chain up for a long time.

It should be possible to have the time to commit scale at least linearly (`O(N)`) with the number of pages written to, if even with a cost involved when deleting commits. Since [`crumbles` was introduced](https://github.com/dusk-network/piecrust/pull/257), `squash_commit` was removed.

#### Possible Solution

This might be a good time to introduce the use of a database such as [`rocksdb`](https://docs.rs/rocksdb/latest/rocksdb/), especially if there's some difficulty optimizing the retrieval of pages the squashing of some commits into another.

The current content of a page can be retrieved by following the commit path until the latest contents are retrieved.

HDauven:
**Consideration**: This is not a change within the protocol, but the state needs to be migrated between the different approaches.

PR #387, opened 2024-09-27
Changed state representation to delta representation to improve commit performance

miloszm:
Changed the state persistence representation into a delta-based representation in order to improve the commit performance.

Commits now store only what is changed, with the rest of content taken either from the base commit or from the current main commit.

Addresses issue #342

### DIFF
--- a/piecrust/src/store.rs
+++ b/piecrust/src/store.rs
@@ -315,7 +315,7 @@ fn commit_from_dir<P: AsRef<Path>>(
         for page_index in &contract_index.page_indices {
             let main_page_path = page_path(&contract_memory_dir, *page_index);
             if !main_page_path.is_file() {
-                let path = ContractSession::do_find_page(
+                let path = ContractSession::find_page(
                     *page_index,
                     maybe_hash,
                     contract_memory_dir.clone(),
--- a/piecrust/src/store.rs
+++ b/piecrust/src/store.rs
@@ -37,6 +37,7 @@ const MEMORY_DIR: &str = "memory";
 const INDEX_FILE: &str = "index";
 const OBJECTCODE_EXTENSION: &str = "a";
 const METADATA_EXTENSION: &str = "m";
+const MAIN_DIR: &str = "main";
 
 /// A store for all contract commits.
 pub struct ContractStore {
@@ -132,6 +133,16 @@ impl ContractStore {
         self.call_with_replier(|replier| Call::CommitDelete { commit, replier })
     }
 
+    /// Finalizes commit
+    ///
+    /// The commit will become a "current" commit
+    pub fn finalize_commit(&self, commit: Hash) -> io::Result<()> {
+        self.call_with_replier(|replier| Call::CommitFinalize {
+            commit,
+            replier,
+        })
+    }
+
     /// Return the handle to the thread running the store's synchronization
     /// loop.
     pub fn sync_loop(&self) -> &thread::Thread {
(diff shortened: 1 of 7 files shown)


piecrust/src/store/session.rs
```diff
@@ -157,6 +158,41 @@ impl ContractSession {
             .map(|c| *c.index.root())
     }
 
+    /// Returns path to a file representing a given commit and page.
+    ///
+    /// Requires a contract's memory path and a main state path.
+    /// Progresses recursively via bases of commits.
+    pub fn do_find_page(
```
moCello, line 165: isn't `find_page` a nicer name for this function?

ureeves: Ditto on `find_page`

piecrust/src/store.rs
```diff
@@ -200,17 +224,65 @@ fn page_path<P: AsRef<Path>>(memory_dir: P, page_index: usize) -> PathBuf {
...
+    let dir = main_dir.as_ref().join(commit_id);
+    fs::create_dir_all(&dir)?;
+    Ok(dir.join(INDEX_FILE))
+}
+
+fn commit_id_to_hash<S: AsRef<str>>(commit_id: S) -> Hash {
+    let hash: [u8; 32] = hex::decode(commit_id.as_ref())
+        .expect("Hex decoding of commit id string should succeed")
+        .try_into()
+        .expect("Commit id string conversion should succeed");
```
ureeves, line 252: I understand that the `expect`s failing here means that the state is corrupt somehow. Any thoughts on how to deal with such a situation in general?

piecrust/src/store.rs
```diff
@@ -200,17 +224,65 @@ fn page_path<P: AsRef<Path>>(memory_dir: P, page_index: usize) -> PathBuf {
...
+
 fn commit_from_dir<P: AsRef<Path>>(
     engine: &Engine,
     dir: P,
 ) -> io::Result<Commit> {
     let dir = dir.as_ref();
+    let mut commit_id: Option<String> = None;
+    let main_dir = if dir
+        .file_name()
+        .expect("Filename or folder name should exist")
```
ureeves, line 264: Ditto on the `expect` here

piecrust/src/store.rs
```diff
@@ -417,179 +527,101 @@ fn write_commit<P: AsRef<Path>>(
...
-                fs::hard_link(base_metadata_path, metadata_path)?;
-
-                for page_index in &elem.page_indices {
-                    let new_page_path = page_path(&memory_dir, *page_index);
-                    let base_page_path =
-                        page_path(&base_memory_dir, *page_index);
-
-                    fs::hard_link(base_page_path, new_page_path)?;
-                }
-            }
```
ureeves, line 588: Very nice on this removal! The power of just structuring data differently based on different assumptions is insane.

ureeves (approved):
I like it. Very nice and focused. I would prefer to not have so many `expect`s around, but that's just consequence of state corruption being possible.

ureeves:
> I like it. Very nice and focused. I would prefer to not have so many `expect`s around, but that's just consequence of state corruption being possible.

Also, please make sure to update the CHANGELOG, and ensure that the commits are prefixed with `piecrust: `

piecrust/tests/root.rs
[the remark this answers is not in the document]
ureeves: I popped your "adjusted tests" commit, and it still passes for me :confused:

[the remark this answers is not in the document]
miloszm: Maybe it is processor type dependent..

piecrust/src/store/session.rs
miloszm: I like adding `do` to names :) as there are so many find_pages around, but I agree with your remarks - I will change that.

piecrust/src/store.rs
miloszm: I agree that we should have a sweep of all `expects` from the point of view of what kind of state corruption do we want to tolerate and how. This is a non-trivial subject so I'd rather leave it as a todo for the near future.

PR #398, opened 2024-10-18
piecrust: improve bootstrap

herr-seppia:
Remove `merkle.root()` calculation during bootstrap

Additionally:
- Add logs for new vm instantiation
- remove useless string clone
- move "calculating root" logs


miloszm (approved):
Looks very good to me, nice changes

Issue #405, opened 2024-11-25
Access to callstack via uplink API


miloszm:
#### Summary
For a new feature of stake contract it is needed to have access to the entire call stack from a called contract method, currently, only the direct caller contract id is available, we need all contract ids.

#### Possible solution design or implementation
Add an uplink API function which will provide an iterator over all contract ids present in the current call stack.

#### Additional context
Required by the stake contract.

PR #406, opened 2024-11-25
Access to callstack via uplink API

miloszm:
Implements functionality which enables access to callstack via uplink API

Addresses issue #405


herr-seppia:
Can this be closed since it's already part of a different PR already merged?

PR #413, opened 2024-12-06
Added serde feature

d-sonuga:
For dusk-network/rusk/issues/2773 and #414

### DIFF
--- a/piecrust-uplink/src/serde_support.rs
+++ b/piecrust-uplink/src/serde_support.rs
@@ -1,0 +1,140 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use alloc::format;
+use alloc::string::String;
+
+use base64::engine::general_purpose::STANDARD as BASE64_STANDARD;
+use base64::Engine;
+use serde::de::{Error as SerdeError, MapAccess, Visitor};
+use serde::ser::SerializeStruct;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+use crate::{ContractId, Event, CONTRACT_ID_BYTES};
+
+impl Serialize for ContractId {
+    fn serialize<S: Serializer>(
+        &self,
+        serializer: S,
+    ) -> Result<S::Ok, S::Error> {
+        let s = hex::encode(self.to_bytes());
+        serializer.serialize_str(&s)
+    }
+}
+
+impl<'de> Deserialize<'de> for ContractId {
+    fn deserialize<D: Deserializer<'de>>(
+        deserializer: D,
+    ) -> Result<Self, D::Error> {
+        let s = String::deserialize(deserializer)?;
+        let decoded = hex::decode(&s).map_err(SerdeError::custom)?;
+        let decoded_len = decoded.len();
+        let byte_length_str = format!("{CONTRACT_ID_BYTES}");
+        let bytes: [u8; CONTRACT_ID_BYTES] =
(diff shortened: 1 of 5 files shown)


piecrust-uplink/src/serde_support.rs
```diff
@@ -0,0 +1,137 @@
...
+
+use alloc::format;
+use alloc::string::String;
+
+use base64::{engine::general_purpose::STANDARD as BASE64_STANDARD, Engine};
+use serde::{
+    de::{self, Error, MapAccess, Visitor},
+    ser::SerializeStruct,
+    Deserialize, Deserializer, Serialize, Serializer,
+};
```
Neotamandua, line 15:
```suggestion
use base64::Engine;
use base64::engine::general_purpose::STANDARD as BASE64_STANDARD;
use serde::de::{self, Error, MapAccess, Visitor};
use serde::ser::SerializeStruct;
use serde::{Deserialize, Deserializer, Serialize, Serializer};
```

piecrust-uplink/tests/serde.rs
```diff
@@ -0,0 +1,45 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+#[cfg(feature = "serde")]
+mod serde {
+    use piecrust_uplink::{ContractId, Event, CONTRACT_ID_BYTES};
+    use rand::{rngs::StdRng, RngCore, SeedableRng};
```
Neotamandua, line 10: as above, imports should be separated by module

Neotamandua (commented):
Looks good too. I just have the same comment as in another PR, whether we want to introduce so many custom error strings or just pass the underlying error through where applicable.

Neotamandua (commented):
Looks good in general. I would suggest using the same seed for randomness in the test cases.

Additionally, I think adding another test that checks the equality of a serialized struct against a serde_json string of an event is a good idea to make sure we have a test case that can catch if the attribute names of the struct ever change.

Example:
```rust
    #[test]
    fn event_fields() {
        let serde_json_string = "{\"source\":\"0000000000000000000000000000000000000000000000000000000000000000\",\"topic\":\"\",\"data\":\"\"}";
        let event = Event {
            source: ContractId::from_bytes([0; CONTRACT_ID_BYTES]),
            topic: String::new(),
            data: Vec::new(),
        };
        let ser = serde_json::to_string(&event).unwrap();
        assert_eq!(serde_json_string, ser);
    }
```

piecrust-uplink/tests/serde.rs
```diff
@@ -0,0 +1,85 @@
...
+    let length_31_enc =
+        "\"1751c37a1dca7aa4c048fcc6177194243edc3637bae042e167e4285945e046\"";
+
+    let contract_id: Result<ContractId, _> =
+        serde_json::from_str(&length_31_enc);
+    assert!(contract_id.is_err());
+}
+
+#[test]
+fn event_fields() {
```
miloszm, line 76: could you rename test functions so that they all have a prefix like serde or sth like that,
function name `contract_id` is misleading and also, it would be nice to be able to run `cargo t serde` in order
to run all serde related tests

miloszm (changes requested):
Overall looks very good, just test fun rename seems important to me
for practical test usability

piecrust-uplink/tests/serde.rs
d-sonuga: The name of the file is serde, so you already have a prefix. To run all serde tests, you can do `cargo t --test serde --features serde --release`

miloszm (approved):
LGTM

Neotamandua (approved):

Neotamandua (approved):

PR #429, opened 2025-02-24
Piecrust: Bump to 0.28.0

moCello:
### Changed

- Update uplink to v0.18.0

[0.28.0]: https://github.com/dusk-network/piecrust/compare/piecrust-0.27.2...piecrust-0.28.0


moCello:
Initially added this PR to release piecrust 0.28 after 0.27.3 was yanked. But this would actually break semver since there are already release candidates released for 0.28.0.

I advice to release 0.28.0 with the changes of the release candidates when we need them.

PR #432, opened 2025-03-10
Upgrade toolchain

d-sonuga:
For https://github.com/dusk-network/rusk/issues/3470.

The rust version chosen is a 1.82.0 nightly (2024-07-30) because it uses LLVM 18. From LLVM 19, the reference-types feature is enabled by default, and I thought it would be a good idea to pick a toolchain that will require the least amount of changes.


Neotamandua (approved):
LGTM. Now that 0.28.0 is out, we should be able to merge this, or? @miloszm  @herr-seppia

Issue #433, opened 2025-03-13
change init to only be called during deploy


miloszm:
#### Summary
'init' method should only be called at contract's deployment time and calling it at all other times should be blocked

#### Possible solution design or implementation
The low level method calling functionality should check called method name and if it is 'init', it should throw an error

PR #434, opened 2025-03-13
blocking 'init' method to be called only during deployment

miloszm:
Functionality and test for blocking 'init' method to be called only during deployment.

Implements issue #433 

It will have a companion test in Rusk as well.

### DIFF
--- a/piecrust/tests/initializer.rs
+++ b/piecrust/tests/initializer.rs
@@ -40,12 +40,6 @@ fn init() -> Result<(), Error> {
         result.is_err(),
         "calling init directly as transaction should not be allowed"
     );
-    // we should not be able to call init as query neither
-    let result = session.call::<u8, ()>(id, CONTRACT_INIT_METHOD, &0xaa, LIMIT);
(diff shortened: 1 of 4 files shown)


piecrust/tests/initializer.rs
```diff
@@ -40,12 +40,6 @@ fn init() -> Result<(), Error> {
         result.is_err(),
         "calling init directly as transaction should not be allowed"
     );
-    // we should not be able to call init as query neither
-    let result = session.call::<u8, ()>(id, CONTRACT_INIT_METHOD, &0xaa, LIMIT);
-    assert!(
-        result.is_err(),
-        "calling init directly as query should not be allowed"
-    );
```
herr-seppia, line 48: Why this has been removed?

miloszm: Because it was duplicated - exactly the same call twice, probably because in the past queries differed from calls and these tests were different but after query removal they became the same

herr-seppia (approved):

Issue #440, opened 2025-04-07
Limit commit object finalization strip-off


miloszm:
#### Summary
After finalization, commit objects need to be stripped-off off the finalized inidices(by index we mean
an element, i.e., a subtree of memory pages for a given contract), otherwise commit grow in size
like snow balls (since a new commit is always created off of a base commit). Sometimes yet such strip-off
is too aggressive, as sometimes this information is still needed. In such case, a check needs to be done 
if a given element is in main, and if it is, it should be recalled back.

#### Possible solution design or implementation
Check if missing element is in main, and if it is, bring it back into the commit.

#### Additional context
This feature is critical for correct root calculations in cases when root is called after finalization and a given element has already beed stripped-off from the commit object.

PR #441, opened 2025-04-07
commits' strip-off algorithm checking if needed elements are in main

miloszm:
commits' strip-off algorithm checking if needed elements are in main

addresses issue #440


herr-seppia (approved):
LGTM

PR #472, opened 2026-06-18
Expose caller stack to call hooks

herr-seppia:
## Summary

- Pass the active caller stack to call hooks before each inter-contract call.
- Keep the stack borrowed as `&[&ContractId]` so hook callers do not copy contract IDs.
- Extend call-hook tests to assert the observed stack for direct and recursive inter-contract calls.
- Document the call-hook API change in the changelog.

## Validation

- `make test`

### DIFF
--- a/piecrust/src/session.rs
+++ b/piecrust/src/session.rs
@@ -93,6 +93,10 @@ impl Drop for Session {
 /// Receives the callee contract ID, the function name, the raw argument bytes,
 /// and the current call stack before the callee is pushed. Returns `Ok(())` to
 /// allow the call, or `Err(reason)` to reject it with a descriptive message.
+///
+/// The callee is not included in the stack because the hook runs before the
+/// callee is pushed. The stack is ordered like `abi::callstack()`: index 0 is
+/// the immediate caller, and the last element is the root contract call.
 #[cfg(feature = "call-hook")]
 pub type CallHook = Box<
     dyn Fn(&ContractId, &str, &[u8], &[&ContractId]) -> Result<(), String>
--- a/piecrust/src/session.rs
+++ b/piecrust/src/session.rs
@@ -90,12 +90,15 @@ impl Drop for Session {
 
 /// A hook called before each inter-contract call.
 ///
-/// Receives the callee contract ID, the function name, and the raw argument
-/// bytes. Returns `Ok(())` to allow the call, or `Err(reason)` to reject it
-/// with a descriptive message.
+/// Receives the callee contract ID, the function name, the raw argument bytes,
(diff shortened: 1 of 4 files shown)


HDauven (approved):

piecrust/src/session.rs
```diff
@@ -90,12 +90,15 @@ impl Drop for Session {
 
 /// A hook called before each inter-contract call.
 ///
-/// Receives the callee contract ID, the function name, and the raw argument
-/// bytes. Returns `Ok(())` to allow the call, or `Err(reason)` to reject it
-/// with a descriptive message.
+/// Receives the callee contract ID, the function name, the raw argument bytes,
+/// and the current call stack before the callee is pushed. Returns `Ok(())` to
+/// allow the call, or `Err(reason)` to reject it with a descriptive message.
```
moCello, line 95: **WARNING:** Public hook contract leaves stack ordering unspecified, and no test pins it — a future `call_ids()` reorder breaks consumers silently with green tests.

**Context:** The ordering ([0] = immediate caller, [n-1] = root) is documented only at `call_tree.rs:362-367`, which is `pub(crate)` and invisible to hook implementers. No test distinguishes order: single-ICC tests assert one-element stacks, and `call_hook_can_deserialize_fn_args` asserts only `len` plus that every frame equals `center_id` — so a top-first vs root-first swap stays green.

**Fix:** State in this doc that index 0 is the immediate caller and the last element is the root. Pin it with a distinct-ID test: deploy two `callcenter`s plus `counter` and delegate through both so the innermost hook observes a stack of distinct IDs, asserting the order.

moCello (commented):
COMMENT — Clean, minimal, correctly scoped change with sound borrow timing and good coverage of stack growth. One item worth addressing: document the call-stack ordering in the public hook doc (index 0 = immediate caller, last = root) and pin it with a distinct-ID test, since the order is the whole point of the feature yet no test distinguishes it.

piecrust/src/session.rs
herr-seppia: You’re right. I checked the full path and my note about reversing was wrong: `abi::callstack()` does not reverse the stack. It skips the current contract and preserves immediate-caller-first ordering.

I’ll document that ordering both on `CallHook` and on the existing `abi::callstack()` API, then add a distinct-ID nested ICC test so the hook ordering is pinned.

moCello (approved):
APPROVE — Correct, minimal, well-tested; the stack ordering matches `abi::callstack()` and is now pinned by a distinct-ID nested test, and prior ordering-doc feedback is fully addressed.

PR #474, opened 2026-06-22
Contain host query panics

d-sonuga:
For https://github.com/dusk-network/piecrust-private/issues/31

Catch host-query pricing/execution panics and return contract panic errors.

Adds a session builder flag and regression coverage for contained and propagated behavior.

### DIFF
--- a/piecrust/CHANGELOG.md
+++ b/piecrust/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `SessionDataBuilder::contain_hq_panics` opt-out for callers that need host-query panics to propagate [#476]
+
+### Changed
+
+- Contain host-query panics by default and return them as contract panic errors [#476]
+
 ## [0.32.0] - 2026-06-19
 
 ### Changed
@@ -564,6 +572,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- ISSUES -->
 [#466]: https://github.com/dusk-network/piecrust/issues/466
+[#476]: https://github.com/dusk-network/piecrust/issues/476
 
 [#rusk_3341]: https://github.com/dusk-network/rusk/issues/3341
 [#440]: https://github.com/dusk-network/piecrust/issues/440
--- a/piecrust/src/imports.rs
+++ b/piecrust/src/imports.rs
@@ -8,6 +8,7 @@ mod wasm32;
 mod wasm64;
 
 use std::any::Any;
+use std::panic::{AssertUnwindSafe, catch_unwind};
 use std::sync::Arc;
 
 use dusk_wasmtime::{
@@ -24,6 +25,7 @@ use crate::instance::{Env, WrappedInstance};
 use crate::session::INIT_METHOD;
 
 pub const GAS_PASS_PCT: u64 = 93;
+const HOST_QUERY_PANICKED: &str = "host query panicked";
 
 pub(crate) struct Imports;
 
@@ -176,14 +178,18 @@ pub(crate) fn hq(
         .host_query_arc(&name)
         .ok_or_else(move || Error::MissingHostQuery(name))?;
     let mut query_arg: Box<dyn Any> = Box::new(());
+    let contain_hq_panics = env.contain_hq_panics();
     let query_cost = {
         let instance = env.self_instance();
-        instance.with_arg_buf(|arg_buf| {
-            host_query.deserialize_and_price(
-                &arg_buf[..arg_len as usize],
-                &mut query_arg,
-            )
-        })
+        let result = instance.with_arg_buf(|arg_buf| {
+            maybe_contain_hq_panic(contain_hq_panics, || {
+                host_query.deserialize_and_price(
+                    &arg_buf[..arg_len as usize],
+                    &mut query_arg,
+                )
+            })
+        });
+        result?
     };
 
     if gas_remaining < query_cost {
@@ -194,8 +200,38 @@ pub(crate) fn hq(
     let instance = env.self_instance();
     instance.set_remaining_gas(gas_remaining - query_cost);
 
-    Ok(instance
-        .with_arg_buf_mut(|arg_buf| host_query.execute(&query_arg, arg_buf)))
+    Ok(instance.with_arg_buf_mut(|arg_buf| {
+        maybe_contain_hq_panic(contain_hq_panics, || {
+            host_query.execute(&query_arg, arg_buf)
+        })
+    })?)
+}
+
+fn maybe_contain_hq_panic<R, F>(contain: bool, f: F) -> Result<R, Error>
+where
+    F: FnOnce() -> R,
+{
+    if !contain {
+        return Ok(f());
+    }
+
+    catch_unwind(AssertUnwindSafe(f)).map_err(|payload| {
+        tracing::error!(
+            panic = %hq_panic_message(payload),
+            "host query panicked"
+        );
+        Error::Panic(HOST_QUERY_PANICKED.to_string())
+    })
+}
+
+fn hq_panic_message(payload: Box<dyn Any + Send>) -> String {
+    match payload.downcast::<String>() {
+        Ok(msg) => *msg,
+        Err(payload) => match payload.downcast::<&'static str>() {
+            Ok(msg) => (*msg).to_string(),
+            Err(_) => "host query panicked".to_string(),
+        },
+    }
 }
 
 pub(crate) fn hd(
--- a/piecrust/src/lib.rs
+++ b/piecrust/src/lib.rs
@@ -140,6 +140,6 @@ pub use error::Error;
 pub use piecrust_uplink::*;
 #[cfg(feature = "call-hook")]
 pub use session::CallHook;
-pub use session::{CallReceipt, Session, SessionData};
+pub use session::{CallReceipt, Session, SessionData, SessionDataBuilder};
 pub use store::PageOpening;
 pub use vm::{HostQuery, VM};
--- a/piecrust/src/session.rs
+++ b/piecrust/src/session.rs
@@ -752,6 +752,10 @@ impl Session {
         self.inner().host_queries.get_arc(name)
     }
 
+    pub(crate) fn contain_hq_panics(&self) -> bool {
+        self.inner().data.contain_hq_panics
+    }
+
     pub(crate) fn nth_from_top(&self, n: usize) -> Option<CallTreeElem> {
         self.inner().call_tree.nth_parent(n)
     }
@@ -1076,11 +1080,18 @@ impl CallReceipt<Vec<u8>> {
     }
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct SessionData {
     data: BTreeMap<Cow<'static, str>, Vec<u8>>,
     pub base: Option<[u8; 32]>,
     excluded_host_queries: BTreeSet<String>,
+    contain_hq_panics: bool,
+}
+
+impl Default for SessionData {
+    fn default() -> Self {
+        Self::builder().into()
+    }
 }
 
 impl SessionData {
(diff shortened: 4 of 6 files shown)


piecrust/src/imports.rs
```diff
@@ -194,8 +199,33 @@ pub(crate) fn hq(
...
+fn maybe_contain_hq_panic<R, F>(contain: bool, f: F) -> Result<R, Error>
+where
+    F: FnOnce() -> R,
+{
+    if !contain {
+        return Ok(f());
+    }
+
+    catch_unwind(AssertUnwindSafe(f))
+        .map_err(|payload| Error::Panic(hq_panic_message(payload)))
```
moCello: **BLOCKER:** A host-query panic message is copied verbatim into contract-readable state; since it is built by native node code it can differ across nodes and split consensus.

**Context:** The case is a nested call: contract A calls contract B, and a host query *inside* B panics. This PR turns that panic into `Error::Panic(msg)` and hands `msg` back up to A, which can read it and persist it on-chain. The path: `callee.call` → `Error::normalize` (recovers `Error::Panic`) → `write_contract_error` → `ContractError::to_parts` writes the full string into A's arg buffer (imports.rs:382).

The message is built by the **host query** — native `std` node code (the `HostQuery` impl in rusk, `vm.rs:346`), *not* the no_std wasm contract. A normal wasm contract panic through this same channel is safe because wasm is deterministic; a native host-query panic is not guaranteed to produce identical text on every node. Only this nested A→B path is observable — a direct contract→host-query panic just returns a `Result` to the node.

**Impact:** If two nodes produce different message bytes, they reach different contract state — a consensus split. This is latent, not immediate: ordinary panics (`unwrap`, index out of bounds, a fixed `expect("...")`) produce identical text on identical binaries. It only diverges when a host query panics with a message that embeds something process-specific:
- a `HashMap`/`HashSet` printed with `{:?}` — iteration order uses a per-process random seed, so it differs even on identical hardware;
- a raw pointer (`{:p}`) — differs via ASLR;
- OS- or locale-dependent error text.

No current rusk host query is known to do this, so a live trigger is unlikely today.

**Severity-rationale:** Blocker despite the low trigger probability: a consensus VM must not contain *any* path by which non-deterministic bytes can reach contract-observable state. The bar is "provably deterministic," not "probably deterministic" — a latent fork vector is unacceptable to merge, and the fix is trivial.

**Fix:** Split the panic detail by audience — the divergent text must not reach the contract, but it need not be lost:
- **Contract (consensus-observable):** hand back a fixed constant message (e.g. `"host query panicked"`) so every node produces identical bytes. To keep a usable signal, give host-query panics their own fixed `ContractError` code (distinct from a callee-contract `Panic(-1)`) rather than just a bare string — a fixed discriminant is deterministic, the message text is not.
- **Host-query impl / node (off-chain):** the impl can still log the real payload and whatever caused the panic, host-side. Full debugging detail is preserved; it simply never enters consensus state.

This removes the divergence class regardless of how any host query writes its panics. For granular reasons a contract should branch on, the durable answer is a follow-up: have host queries return *typed* errors mapped to stable codes for expected failure modes, reserving the panic path for genuine bugs (whose message is inherently non-deterministic). Note: the two containment tests assert the exact raw panic string, so they'd switch to asserting the constant; the nested A→B path is untested today.

piecrust/src/session.rs
```diff
@@ -1148,11 +1165,17 @@ impl SessionDataBuilder {
         self
     }
 
+    pub fn contain_hq_panics(mut self, contain: bool) -> Self {
+        self.contain_hq_panics = contain;
+        self
+    }
```
moCello, line 1172: **WARNING:** Public-API addition plus a default-behavior flip in a consensus-relevant VM, with no linked tracking issue.

**Context:** This PR adds public surface (`SessionDataBuilder::contain_hq_panics`, `SessionData::contain_hq_panics`, the `SessionDataBuilder` re-export in lib.rs) and flips default execution semantics: host-query panics that previously unwound the VM now surface as `Error::Panic` by default. A change at this altitude wants a tracking issue for rationale and downstream coordination (rusk/vm, rusk/contracts).

**Fix:** Open an issue describing the motivation and the default flip, and link this PR to it. Maintainer judgment call on whether the descriptive title alone is sufficient.

piecrust/tests/host.rs
```diff
@@ -89,17 +123,87 @@ fn new_ephemeral_vm() -> Result<VM, Error> {
...
+    Ok(())
+}
+
+#[test]
+pub fn host_query_panic_propagates_when_containment_is_disabled()
+-> Result<(), Error> {
+    let mut vm = VM::ephemeral()?;
+    vm.register_host_query("very_expensive", PanicsInExecuteQuery);
+    let mut session =
+        vm.session(SessionData::builder().contain_hq_panics(false))?;
```
moCello, line 224: **NOTE:** The `contain_hq_panics(false)` opt-out has no in-tree caller other than this propagation test, and it re-enables the unwind-out-of-the-VM behavior the rest of the PR exists to prevent.

**Fix:** Confirm an embedder genuinely needs the propagate path; if not, dropping the opt-out shrinks the API surface. If kept, document on the builder method that disabling containment lets a host-query panic unwind through the VM and abort the calling stack.

moCello (changes requested):
REQUEST CHANGES — Clean, well-tested change with both prior review comments resolved, but one path must be closed before merge: a contained host-query panic forwards its raw message into contract-observable state (nested A→B call path), and because that message is built by native node code it can differ across nodes — a non-deterministic input to consensus state. A consensus VM must not carry any such path regardless of how unlikely a live trigger is today; map the contained panic to a fixed constant and log the real payload instead (trivial, and it removes the class entirely). Separately, the public-API addition + default-behavior flip ships with no linked issue — consider opening one for rusk/vm + rusk/contracts coordination (maintainer call).

piecrust/tests/host.rs
d-sonuga: I added this option to disable it to preserve historical behavior, in case Rusk needs it for historical execution.

piecrust/src/lib.rs
```diff
@@ -140,6 +140,6 @@ pub use error::Error;
 pub use piecrust_uplink::*;
 #[cfg(feature = "call-hook")]
 pub use session::CallHook;
-pub use session::{CallReceipt, Session, SessionData};
+pub use session::{CallReceipt, Session, SessionData, SessionDataBuilder};
```
moCello, line 143: **WARNING:** The `SessionDataBuilder` re-export and the `SessionData::contain_hq_panics()` accessor add public surface with no consumer — in piecrust or rusk.

**Context:** Only the `contain_hq_panics(bool)` opt-out is load-bearing (rusk needs propagate for faithful historical replay).
- Re-export of `SessionDataBuilder`: nothing names the type — rusk holds the builder by inference (dusk-vm `vm/src/lib.rs:225`, imports only `{SessionData, VM}`).
- Public accessor `contain_hq_panics`: read only by piecrust tests; production uses `pub(crate)` `Session::contain_hq_panics()` (`imports.rs:181`), rusk never reads it.

`#476` justifies the opt-out but is silent on the two other additions to the public API. Round 1 only required the issue link (added) — this is the minimality follow-up.

**Fix:** Make the accessor `pub(crate)` (or drop it, asserting the default in a unit test); drop the re-export unless rusk will name the type.

piecrust/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Contain host-query panics by default and return them as contract panic errors [#476]
```
moCello, line 16: **NOTE:** Changelog records the default-behavior flip (`### Changed`) but not the new public API it ships — the `contain_hq_panics(bool)` opt-out is a callable a downstream consumer can now use.

**Context:** Under keep-a-changelog a new public callable belongs under `### Added`, distinct from the `### Changed` behavior line. Document only what stays public: the `SessionDataBuilder` re-export and `SessionData::contain_hq_panics()` accessor are pending the minimality WARNING above — add an `### Added` line for them only if they survive that decision; if narrowed/dropped, there's nothing to record (and nothing to document is the cleaner outcome).

**Fix:** Add an `### Added` bullet for the `contain_hq_panics` opt-out; reconcile any re-export/accessor line with the minimality decision rather than documenting surface that may be removed.

contracts/double_counter/src/lib.rs
```diff
@@ -55,6 +55,14 @@ impl DoubleCounter {
...
         uplink::call(contract, "hello", &())
     }
+
+    /// Call the host-query test contract and return its error to the caller.
+    pub fn delegate_host_query(
+        &mut self,
+        contract: ContractId,
+    ) -> Result<(), ContractError> {
+        uplink::call(contract, "host_very_expensive", &())
+    }
```
moCello: **NOTE:** This `contracts/` fixture and the `piecrust/` change ship in one commit whose subject lowercases its first word — both diverge from repo commit convention.

**Context:** Convention is one commit per crate per concern (`contracts` is its own scope, ordered before `piecrust`) with the subject's first word capitalized. No functional impact, and the fixture genuinely supports the new nested host-query test — a fair argument for keeping them together.

**Fix:** Optionally split `double_counter` into a `contracts:`-scoped commit ordered before the `piecrust:` commit, and capitalize the subject (`Contain ...`). Maintainer judgment call on whether the test-fixture coupling justifies one commit.

moCello (commented):
COMMENT — The prior blocker is correctly closed: a contained host-query panic now hands the contract a fixed constant while the real payload is logged off-chain only, and the nested A→B regression test proves the contract-observable message is deterministic. The containment wrap covers the sole contract→host call path and is placed so the arg-buffer borrow releases cleanly on panic; gas accounting on the execute path is correct. Default-on behavior, changelog, issue linkage, and opt-out docs are all in place.
One thing worth narrowing before merge: of the three new public items, only the `contain_hq_panics(bool)` opt-out has a consumer (rusk's historical-replay path) — the `SessionDataBuilder` re-export and the `SessionData::contain_hq_panics()` accessor have no in-tree or rusk caller, so make the accessor `pub(crate)` and drop the re-export unless rusk will name the type. (Round 1 flagged this surface but only required a linked issue, which you added — this is the minimality follow-up.) Two minor cleanups beyond that: a changelog `### Added` line for the new opt-out API, and commit hygiene (split the `contracts/` fixture and capitalize the subject). The earlier idea of a distinct error code for host-query panics (vs reusing `Panic`) is a reasonable follow-up but not required for correctness.

piecrust/src/lib.rs
d-sonuga: I exported that so that it can be used downstream like this https://github.com/dusk-network/rusk/compare/master...contain-hq-panic. To avoid having to modify the existing `vm.session` function that takes a fixed number of arguments and is used in multiple places.  The idea is that something else can just take the builder with the default arguments and modify it directly.

piecrust/src/session.rs
```diff
@@ -1113,6 +1125,10 @@ impl SessionData {
     pub fn excluded_host_queries(&self) -> Iter<String> {
         self.excluded_host_queries.iter()
     }
+
+    pub fn contain_hq_panics(&self) -> bool {
```
moCello: **NOTE:** Public getter has no caller — only the test reads it; production reads the flag internally via `Session::contain_hq_panics` field access, not this method.

**Context:** The look-alike `pub fn excluded_host_queries()` one line up is genuinely consumed (vm.rs:233 applies the exclusions), so it earns `pub`. This getter is the only `SessionData` field getter with no consumer — public-surface bloat, not a correctness issue.

**Fix:** Optional — narrow to `pub(crate)` and move the default-on assertion into an in-crate unit test, or keep it if a downstream reader is planned.

moCello (approved):
APPROVE — The round-2 determinism blocker is properly closed: a contained host-query panic hands the contract a fixed constant while the real payload is logged off-chain only, and the nested A→B test proves the contract-observable message is deterministic. Default-on containment, the opt-out for historical replay, changelog, and issue linkage are all in place. One optional NOTE remains — a public getter with no caller, the flag being read internally via the field — and it does not block merge.

PR #483, opened 2026-06-25
piecrust: Deduplicate stored contract payloads

### LINKED ISSUE #470: storage: avoid duplicating identical contract payloads
#### Summary

Repeated deployments of contracts with identical WASM currently store duplicate contract payloads on disk. This makes persisted state grow linearly with duplicate deployments, even when the bytecode is identical.

This matters for factories, collections, templates, and other workflows where many contracts share the same code but differ by contract id or initial state.

Store identical contract payloads once and keep deployment-specific references separate.

The solution should preserve existing persisted states and keep newly deployed contracts behaving the same after reopen/restart. If filesystem-level linking is used, it should have a portable fallback for environments where links are unavailable.

HDauven:
Resolves dusk-network/piecrust#470.

Moved from dusk-network/piecrust-private#24.

## Rationale

Repeated deployments of the same contract currently duplicate the raw WASM and compiled module artifacts on disk. This is wasteful for token factories, collections, or any workflow where many contracts share identical bytecode but have different deployment IDs or initial state.

This change stores contract payloads by content hash and links deployment-specific files to the canonical payload. Existing states remain readable because the layout only affects newly written payloads; old files are left where they are. If hard links are not available, the implementation falls back to copying so the storage path remains portable.

The change was validated with regression coverage for duplicate deployments, VM reopen behavior, and local persisted-state testing with historic deployments followed by new duplicate deployments.

## Validation

- `make fmt CHECK=1`
- `cargo check -p piecrust --lib --all-features`
- `git diff --check origin/main..HEAD`
- `make test`
- `make cq`
- `make no-std`

### DIFF
--- a/piecrust/CHANGELOG.md
+++ b/piecrust/CHANGELOG.md
@@ -16,8 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Declare minimum supported Rust version of 1.94.
 - Contain host-query panics by default and return them as contract panic errors [#476]
 - Move `rand` from runtime dependencies to dev-dependencies.
-- Deduplicate stored contract bytecode and compiled module payloads.
-- Remove unreferenced deduplicated contract payload canonicals.
+- Deduplicate stored contract bytecode and compiled module payloads [#470].
+- Remove unreferenced deduplicated contract payload canonicals [#470].
 
 ### Fixed
 
@@ -586,6 +586,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- ISSUES -->
 [#466]: https://github.com/dusk-network/piecrust/issues/466
+[#470]: https://github.com/dusk-network/piecrust/issues/470
 [#476]: https://github.com/dusk-network/piecrust/issues/476
 
 [#rusk_3341]: https://github.com/dusk-network/rusk/issues/3341
--- a/piecrust/CHANGELOG.md
+++ b/piecrust/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Declare minimum supported Rust version of 1.94.
 - Contain host-query panics by default and return them as contract panic errors [#476]
 - Move `rand` from runtime dependencies to dev-dependencies.
+- Deduplicate stored contract bytecode and compiled module payloads.
 
 ### Fixed
 
--- a/piecrust/CHANGELOG.md
+++ b/piecrust/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Contain host-query panics by default and return them as contract panic errors [#476]
 - Move `rand` from runtime dependencies to dev-dependencies.
 - Deduplicate stored contract bytecode and compiled module payloads.
+- Remove unreferenced deduplicated contract payload canonicals.
 
 ### Fixed
 
--- a/piecrust/src/store/dedup.rs
+++ b/piecrust/src/store/dedup.rs
@@ -8,6 +8,7 @@ use std::fs::{self, OpenOptions};
 use std::io::{self, Write};
 use std::os::unix::fs::MetadataExt;
 use std::path::{Path, PathBuf};
+use std::process;
 use std::sync::Mutex;
 use std::sync::atomic::{AtomicU64, Ordering};
 
@@ -58,7 +59,7 @@ where
     replace_with_hard_link_or_copy(&canonical, target)?;
 
     if let Some(old_hash) = old_hash.filter(|old_hash| old_hash != &hash) {
-        remove_unreferenced_hash(store_root, kind, &old_hash)?;
+        remove_unreferenced_hash_best_effort(store_root, kind, &old_hash);
     }
 
     Ok(())
@@ -81,12 +82,26 @@ where
     match fs::remove_file(target) {
         Ok(()) => {
             if let Some(hash) = hash {
-                remove_unreferenced_hash(store_root, kind, &hash)?;
+                remove_unreferenced_hash_best_effort(store_root, kind, &hash);
             }
             Ok(())
         }
         Err(err) if err.kind() == io::ErrorKind::NotFound => Ok(()),
         Err(err) => Err(err),
+    }
+}
+
+fn remove_unreferenced_hash_best_effort(
+    store_root: &Path,
+    kind: Kind,
+    hash: &str,
+) {
+    if let Err(err) = remove_unreferenced_hash(store_root, kind, hash) {
+        tracing::warn!(
+            kind = kind.directory(),
+            hash,
+            "failed to remove unreferenced dedup canonical: {err}"
+        );
     }
 }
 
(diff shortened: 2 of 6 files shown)


piecrust/CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Declare minimum supported Rust version of 1.94.
 - Contain host-query panics by default and return them as contract panic errors [#476]
+- Deduplicate stored contract bytecode and compiled module payloads [#470].
```
moCello, line 19: **WARNING:** `[#470]` referenced twice but never defined — renders as literal text, unlike the working `[#476]`.

**Fix:** add `[#470]: https://github.com/dusk-network/piecrust/issues/470` to the link-def block (bottom of file, by `[#476]`). Consider folding the two bullets into one — store-by-hash and evict-unreferenced are halves of the same feature.

piecrust/src/store/dedup.rs
```diff
@@ -0,0 +1,307 @@
...
+            let result = fs::rename(&tmp, target);
+            if result.is_err() {
+                let _ = fs::remove_file(&tmp);
+            }
+            result
+        }
+        Err(_) => {
+            let _ = fs::remove_file(&tmp);
+            let result = fs::copy(canonical, &tmp)
+                .and_then(|_| fs::rename(&tmp, target));
```
moCello: **WARNING:** the copy fallback — the PR's stated portability guarantee for link-less filesystems — has no test coverage. Every test runs where `hard_link` succeeds, so this arm never executes.

**Context:** the one test touching this function, `hard_link_rename_failure_removes_tmp_link`, hits the `hard_link`-success path (link to tmp works, rename-over-a-directory fails); the `Err(_) => copy` branch is never reached. Correct by inspection, but a regression on a link-less filesystem would surface nowhere.

**Fix:** extract the copy step into a small helper and unit-test it — assert the target is an independent inode with identical bytes, and that eviction afterward leaves the copy intact.

piecrust/src/store/dedup.rs
```diff
@@ -0,0 +1,307 @@
...
+}
+
+fn is_hash(hash: &str) -> bool {
+    hash.len() == HASH_HEX_LEN && hash.bytes().all(|b| b.is_ascii_hexdigit())
+}
+
+fn write_canonical_if_missing(path: &Path, bytes: &[u8]) -> io::Result<()> {
+    match fs::read(path) {
+        Ok(existing) if existing == bytes => Ok(()),
+        Ok(_) => replace_file(path, bytes),
```
moCello, line 154: **NOTE:** every dedup write reads the entire existing canonical and byte-compares it, though the filename already *is* the content hash of `bytes`.

**Impact:** for the motivating workload (redeploy the same large module N times) this is a full read + compare on every deploy. The `Ok(_) =>` mismatch arm is reachable only under on-disk corruption or a blake3 collision — and if it fires while the canonical has live links, `replace_file`'s rename detaches those links onto the stale inode.

**Fix:** decide the tradeoff explicitly. Keep the integrity check but hash the read bytes instead of full-buffer comparison; or treat existence at the hash-named path as sufficient and skip the read, reserving rewrite for the missing case.

piecrust/src/store/dedup.rs
```diff
@@ -0,0 +1,307 @@
...
+    kind: Kind,
+    target: T,
+    bytes: B,
+) -> io::Result<()>
+where
+    P: AsRef<Path>,
+    T: AsRef<Path>,
+    B: AsRef<[u8]>,
+{
+    let _guard = DEDUP_LOCK.lock().unwrap_or_else(|e| e.into_inner());
```
moCello, line 48: **NOTE:** `DEDUP_LOCK` is a single process-global mutex held across hashing, canonical I/O, linking and eviction, so all contract-payload writes across all VM instances in the process serialize here — where the previous `fs::write` path held no lock at all.

**Context:** the lock keeps eviction's read-nlink-then-remove from racing a concurrent link, but its granularity is the whole process rather than the store.

**Fix:** acceptable as-is for correctness; if concurrent-deploy throughput matters, key the lock per `store_root` (a map of per-root locks) instead of one global.

piecrust/src/store/dedup.rs
```diff
@@ -0,0 +1,307 @@
...
+fn tmp_path(path: &Path) -> PathBuf {
+    let counter = TMP_COUNTER.fetch_add(1, Ordering::Relaxed);
+    let file_name = path
+        .file_name()
+        .map(|name| name.to_string_lossy())
+        .unwrap_or_else(|| "dedup".into());
+
+    path.with_file_name(format!(
+        ".{file_name}.dedup-tmp-{}-{counter}",
+        std::process::id()
```
moCello: **NOTE:** fully-qualified call in a function body; no other inline qualified `std::` call exists in `piecrust/src`, and the module imports everything else at the top.

**Fix:** add `use std::process;` at the top and call `process::id()`.

piecrust/src/store/dedup.rs
```diff
@@ -0,0 +1,307 @@
...
+
+    if has_live_hard_link(&canonical)? {
+        return Ok(());
+    }
+
+    fs::remove_file(canonical)
+}
+
+fn has_live_hard_link(path: &Path) -> io::Result<bool> {
+    Ok(fs::metadata(path)?.nlink() > 1)
```
moCello, line 132: **NOTE:** the write-path "keep shared canonical" branch is untested. `removing_duplicate_module_keeps_shared_canonical` covers the keep case via the *remove* flow; `migration_removes_replaced_bytecode_canonical` covers only the single-reference drop-to-1 *evict* case on the write flow. No test redeploys over one of several contracts sharing a bytecode, where eviction must *keep* the canonical because other links remain.

**Context:** separately, the two `*_ignores_*_eviction_failure` tests (dedup.rs:276, dedup.rs:298) force eviction failure via `chmod 0o500` on the canonical dir. Root bypasses that mode, so under root the eviction *succeeds* and `assert!(old_canonical.exists())` / `assert!(canonical.exists())` fail — these tests assume a non-root runner (fine on GH's `runner` user, breaks in root containers).

**Fix:** add a write-path case with two contracts sharing a bytecode, migrate one, assert the canonical survives and the other still resolves. Optionally guard the chmod tests against running as root.

moCello (commented):
COMMENT — no correctness blocker; the dedup layout, nlink refcounting, and eviction are sound across hard-link and copy modes, and the reader/finalize paths are unaffected. Two things warrant attention before merge: the broken `[#470]` CHANGELOG reference (add its link definition), and the absence of any test for the copy fallback even though the PR markets it as the portability guarantee — extract the copy step and unit-test it. The remaining notes (redundant canonical read, process-global lock granularity, inline qualified path, write-path eviction-keep coverage + root-fragile chmod tests) are quality improvements a maintainer can weigh, not gates.

piecrust/src/store/dedup.rs
```diff
@@ -0,0 +1,345 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use std::fs::{self, OpenOptions};
+use std::io::{self, Write};
+use std::os::unix::fs::MetadataExt;
```
Neotamandua, line 9: Unconditional import (missing #[cfg(unix)])

HDauven: Piecrust/Crumbles is Unix-only, so this now keeps the Unix `MetadataExt` dependency explicit instead of pretending to support non-Unix targets.

PR #484, opened 2026-06-25
piecrust: fix deferred commit finalization

d-sonuga:
Fix bug where a pending finalize commit request is executed as a delete commit in contract store.

### DIFF
--- a/piecrust/src/store.rs
+++ b/piecrust/src/store.rs
@@ -270,6 +270,80 @@ pub(crate) enum Call {
     SessionDrop(Hash),
 }
 
+enum DeferredCommitOp {
+    Delete(mpsc::SyncSender<io::Result<()>>),
+    Finalize(mpsc::SyncSender<io::Result<()>>),
+}
+
+fn queue_deferred_op(
+    deferred_ops: &mut BTreeMap<Hash, Vec<DeferredCommitOp>>,
+    root: Hash,
+    op: DeferredCommitOp,
+) {
+    match deferred_ops.entry(root) {
+        Vacant(entry) => {
+            entry.insert(vec![op]);
+        }
+        Occupied(mut entry) => {
+            entry.get_mut().push(op);
+        }
+    }
+}
+
+fn delete_commit(
+    root_dir: &Path,
+    commit_store: &Arc<Mutex<CommitStore>>,
+    root: Hash,
+    replier: mpsc::SyncSender<io::Result<()>>,
+) {
+    let io_result = CommitRemover::remove(root_dir, root);
+    commit_store.lock().unwrap().remove_commit(&root, false);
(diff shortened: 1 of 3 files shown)


piecrust/src/store.rs
```diff
@@ -407,8 +443,8 @@ fn sync_loop<P: AsRef<Path>>(
             }
             // Signal that a session with a base commit has dropped and
             // decrements the hold count, once incremented using
-            // `Call::SessionHold`. If this is the last session that held that
-            // commit, and there are queued deletions, execute them.
+            // `Call::SessionHold`. If this is the last session that held the
```
moCello: **NOTE:** Comment references `Call::SessionHold`, which does not exist — the variant is `Call::CommitHold` (store.rs:425).

**Context:** Same stale name also at store.rs:386 (`held by at least one session using \`Call::SessionHold\``), outside this diff. The PR edits the line-446 comment, so fixing both here is cheap.

**Fix:** Replace `Call::SessionHold` with `Call::CommitHold` in both comments.

piecrust/src/store.rs
```diff
@@ -270,6 +270,73 @@ pub(crate) enum Call {
...
+    if commit_store.get_commit(&root).is_none() {
+        tracing::trace!("finalizing commit finished");
+        let _ = replier.send(Ok(()));
+        return;
+    }
+
+    let io_result = CommitFinalizer::finalize(root, root_dir);
+    commit_store.remove_commit(&root, true);
+    tracing::trace!("finalizing commit finished");
+    let _ = replier.send(io_result);
```
moCello, line 328: **NOTE:** Extraction drops the failure trace. The old inline finalize matched `io_result` and logged the error on `Err`; `finalize_commit` now logs `"finished"` unconditionally, so a finalize IO failure is no longer distinguishable in traces. Error still propagates via the replier, so this is observability-only — but a diagnostic regression in a state-storage zone.

**Fix:** Match on `io_result` and trace the error before sending, as the original did.

moCello (approved):
APPROVE — Correct, minimal fix for a real data-loss bug: a deferred `finalize_commit` was executed as a delete, discarding committed state (and skipping the `main_index` promotion that finalize's deep removal performs) instead of promoting it. The typed-op dispatch is the right shape, and the three added tests cover the direct path plus both held-session deferral paths with filesystem-level assertions. Only two NOTE-level nits remain (a stale `Call::SessionHold` comment name, a lost error-path trace in the extracted finalize helper). Two off-finding asks for a maintainer: capitalize the commit subject to match the repo's `<scope>: <Capitalized>` convention, and — given the data-loss class — consider opening a tracking issue for visibility. Neither blocks.

HDauven (approved):

PR #486, opened 2026-06-30
piecrust: prevent dangling commit index element references

d-sonuga:
Fix undefined behavior where a reference is made from a pointer to a contract index element after the owning mutex guard is dropped.

### DIFF
--- a/piecrust/CHANGELOG.md
+++ b/piecrust/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Keep inherited commit index lookups guard-bound to avoid dangling references when commits are removed concurrently.
 - Preserve queued `finalize_commit` operations when a commit is held by a live session.
 
 ### Removed
--- a/piecrust/Makefile
+++ b/piecrust/Makefile
@@ -3,6 +3,8 @@ help: ## Display this help screen
 
 test: ## Run tests
 	@cargo test --all-features --color=always
+	@rustup component add --toolchain nightly miri
+	@cargo +nightly miri test --all-features --color=always test_no_dangling_commit_index_reference
 
 clippy: ## Run clippy
 	@cargo clippy --all-features -- -D warnings
--- a/piecrust/src/store/commit.rs
+++ b/piecrust/src/store/commit.rs
@@ -10,14 +10,15 @@ pub mod remover;
 pub mod writer;
 
 use std::cell::Ref;
-use std::sync::{Arc, Mutex};
+use std::ops::Deref;
+use std::sync::{Arc, Mutex, MutexGuard};
 
 use piecrust_uplink::ContractId;
 use tracing::debug;
 
 use crate::PageOpening;
 use crate::store::Memory;
-use crate::store::commit_store::CommitStore;
+use crate::store::commit_store::{CommitStore, ElementOwner};
 use crate::store::hasher::Hash;
 use crate::store::index::{ContractIndexElement, NewContractIndex};
 use crate::store::tree::{ContractsMerkle, position_from_contract};
@@ -29,6 +30,38 @@ pub(crate) struct Commit {
     maybe_hash: Option<Hash>,
     commit_store: Option<Arc<Mutex<CommitStore>>>,
     base: Option<Hash>,
+}
+
+/// A reference to a local or store-backed contract index element.
+pub(crate) enum ContractIndexElementRef<'a> {
+    /// A reference to an element in this `Commit`'s `index` map.
+    Local(&'a ContractIndexElement),
+    /// A reference to an element in the `CommitStore` while its guard is held.
+    Store {
+        /// Keeps inherited entries alive while callers read through this ref.
+        guard: MutexGuard<'a, CommitStore>,
+        /// Identifies which store index to read under the guard.
+        owner: ElementOwner,
+        /// Identifies the specific contract entry under the guarded owner.
+        contract_id: ContractId,
+    },
+}
+
+impl Deref for ContractIndexElementRef<'_> {
+    type Target = ContractIndexElement;
+
+    fn deref(&self) -> &Self::Target {
+        match self {
+            Self::Local(element) => element,
+            Self::Store {
+                guard,
+                owner,
+                contract_id,
+            } => guard.get_element(*owner, contract_id).expect(
+                "index element must exist while the store guard is held",
+            ),
+        }
+    }
 }
 
 impl Commit {
@@ -113,9 +146,12 @@ impl Commit {
                 );
             }
         }
-        let (element, contracts_merkle) =
-            self.element_and_merkle_mut(&contract_id);
-        let element = element.unwrap();
+        let (element, contracts_merkle) = (
+            self.index
+                .get_mut(&contract_id)
+                .expect("commit insertion must create a local index element"),
+            &mut self.contracts_merkle,
+        );
 
         element.set_len(memory.current_len());
 
@@ -202,14 +238,16 @@ impl Commit {
     pub fn index_get(
         &self,
         contract_id: &ContractId,
-    ) -> Option<&ContractIndexElement> {
+    ) -> Option<ContractIndexElementRef<'_>> {
+        if let Some(e) = self.index.get(contract_id) {
+            return Some(ContractIndexElementRef::Local(e));
+        }
+
         Hulk::deep_index_get(
-            &self.index,
             *contract_id,
-            self.commit_store.clone(),
+            self.commit_store.as_ref(),
             self.base,
         )
-        .map(|a| unsafe { &*a })
     }
 
     pub fn index(&self) -> &NewContractIndex {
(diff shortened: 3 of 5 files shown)


piecrust/Makefile
```diff
@@ -3,6 +3,7 @@ help: ## Display this help screen
 
 test: ## Run tests
 	@cargo test --all-features --color=always
+	@cargo +nightly miri test --all-features --color=always test_no_dangling_commit_index_reference
```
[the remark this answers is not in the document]
d-sonuga: It's just one test for now. This shouldn't be a problem.

piecrust/Makefile
```diff
@@ -3,6 +3,8 @@ help: ## Display this help screen
 
 test: ## Run tests
 	@cargo test --all-features --color=always
+	@rustup component add --toolchain nightly miri
+	@cargo +nightly miri test --all-features --color=always test_no_dangling_commit_index_reference
```
moCello: **WARNING:** Folding Miri into the default `test` target forces a nightly toolchain and a network `rustup component add` on every `make test`, plus a recompile under the Miri interpreter.

**Impact:** Stable-only or offline environments that currently pass `make test` fail or stall on the component install; the workspace `make test` delegates here, so all contributors hit it.

**Severity-rationale:** Build-tooling friction, not a correctness defect — a maintainer call on whether the regression coverage belongs in the default path.

**Fix:** Move the Miri invocation to a dedicated target (e.g. `miri:`) or gate it behind an env var, and run that from CI rather than from `test`.

piecrust/src/store/commit.rs
```diff
@@ -223,69 +261,92 @@ impl Commit {
...
+        let element = descendant
+            .index_get(&contract_id)
+            .expect("ancestor should contain the element");
+
+        let removed = remove_commit_if_possible(&commit_store, &ancestor_hash);
+
+        // Old code removes the ancestor before this read, so Miri should reject
+        // the stale reference. Fixed code keeps the store locked while
+        // `element` is alive, so the attempted removal is deferred.
+        let _ = element.len();
```
moCello: **NOTE:** Regression test has no assertion — its value rests entirely on Miri's use-after-free detector firing; nothing checks the intended behavior that holding the ref keeps the store locked.

**Context:** In fixed code `remove_commit_if_possible` always observes `WouldBlock` and returns `false` because `element` holds the guard. Asserting that makes the invariant explicit and yields a readable failure if a regression drops the guard, instead of only an opaque Miri diagnostic.

**Fix:** Add `assert!(!removed, "removal must be deferred while the element ref holds the store lock");` after the call.

moCello (commented):
COMMENT — The core fix is correct and minimal: the guard-bound `ContractIndexElementRef` keeps the store locked exactly as long as a caller reads an inherited entry, the borrow-not-clone change to `deep_index_get` makes the returned guard sound, and the unsafe raw-pointer helpers are gone with no leftover references. Lifetimes are compiler-enforced, and the prior review's double-lookup and redundant-Deref comments are addressed. One item needs a maintainer decision: the Makefile change folds Miri into the default `test` target, forcing a nightly toolchain and a network component install on every `make test`. The single note — a missing explicit assertion in the Miri-only regression test — is polish, not a blocker.

piecrust/src/store/commit.rs
d-sonuga: The intended behavior I was going for here is: after ancestor removal is attempted, using the returned element is still valid and not UB. I went with doing it this way because it's possible for that function to be implemented as a copy of the element instead, in which case, it does not have to hold the lock and will still not be UB.

This tests the right contract under Miri: removal may or may not happen, but accessing element's fields after the attempted removal must not be UB.

piecrust/Makefile
d-sonuga: `make fmt` already uses the nightly toolchain so this really isn't a problem.
`rustup component add` will only download it once.

There is no build tooling friction anywhere. It's just `make test`.

piecrust/src/store/commit.rs
```diff
@@ -206,14 +242,16 @@ impl Commit {
...
+
         Hulk::deep_index_get(
-            &self.index,
             *contract_id,
-            self.commit_store.clone(),
+            self.commit_store.as_ref(),
             self.base,
         )
-        .map(|a| unsafe { &*a })
     }
```
moCello: **BLOCKER:** `index_get` can hand back a value that holds the process-wide `CommitStore` mutex — holding it across any store access deadlocks the VM's single commit thread.

**Context:** The `Store` variant owns a `MutexGuard`, the `Local` variant owns no lock, and which one you get depends on whether the contract happens to live in this commit's index or in an ancestor. The call site cannot tell, and a test touching only locally-present contracts never sees the locking path. On the other side of that lock sits `sync_loop`, the single thread servicing `Call::Commit`, `CommitDelete`, `CommitFinalize`, `CommitHold` and `SessionDrop` for every session in the process. Hold the handle across any of those calls and the session thread blocks `sync_loop` on the store lock while waiting for `sync_loop`'s reply — the whole commit pipeline hangs. Two `index_get` calls in one expression self-deadlock outright: `std::sync::Mutex` is not reentrant and the first handle outlives the second call.

**Impact:** All four call sites in `session.rs` drop the handle within their enclosing statement, so nothing hangs today. Nothing in the type or the signature keeps the next caller honest, and the failure mode is a VM-wide hang rather than a panic.

**Fix:** Don't let the guard escape. Bound it to a callback — `fn with_index_element<R>(&self, id: &ContractId, f: impl FnOnce(&ContractIndexElement) -> R) -> Option<R>` — which fits all four callers (three want existence only, one wants `(page_indices, len)`). Returning an owned `ContractIndexElement` clone also works: the session already receives a deep clone of its whole base `Commit`, index and merkle included (`store.rs:217`), so one more element clone is in line with what it already pays.

piecrust/src/store/commit.rs
```diff
@@ -31,6 +32,38 @@ pub(crate) struct Commit {
...
+            Self::Store {
+                guard,
+                owner,
+                contract_id,
+            } => guard.get_element(*owner, contract_id).expect(
+                "index element must exist while the store guard is held",
+            ),
+        }
+    }
+}
```
moCello: **WARNING:** `Deref` re-resolves the element from `(owner, contract_id)` on every call and `expect`s a lookup that cannot fail — a panic in library code standing in for an invariant the compiler could carry.

**Context:** The handle stores coordinates rather than the reference it was built from, so each deref replays two `BTreeMap` lookups. That is why `session.rs:292` needs the `let elem = &*elem;` reborrow to avoid paying it twice — a line the next reader will take for a no-op and delete.

**Fix:** Resolve the element once under the guard, into an owned clone or a callback, per the `index_get` finding: no coordinates left to replay, and no `Option` to unwrap.

piecrust/Makefile
moCello: **WARNING:** The Miri gate hangs on a hardcoded test name — rename the test and the filter matches nothing, `cargo test` exits 0, and the regression gate silently stops guarding anything.

**Context:** The filter cannot simply be dropped either: without `--lib` the command drives all 27 integration test binaries through the interpreter to run zero tests. And folding Miri into the default `test` target puts it on both CI runners (`dusk_ci.yml` runs the root `make test` on Linux and macOS-ARM, which delegates here), so each pays nightly resolution, a networked `rustup component add` with no `|| true` fallback (unlike the root `fmt` target's), and a full Miri recompile on top of the `cargo test` that just ran.

**Fix:** Give Miri its own target invoked from a dedicated CI job, scope it with `--lib`, and make it self-verifying — either drop the name filter (gating wasmtime-dependent unit tests with `#[cfg(not(miri))]`) or fail the target when zero tests ran.

piecrust/CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Reject distinct explicit `ContractId`s that map to an occupied contract Merkle position.
+- Keep inherited commit index lookups guard-bound to avoid dangling references when commits are removed concurrently.
```
moCello: **WARNING:** A soundness fix for undefined behaviour in the state store has no tracking issue, and the entry names the mechanism chosen rather than the defect a consumer cares about.

**Context:** Neighbouring entries carry issue links (`[#476]`). "Guard-bound lookups" is an implementation detail this PR may change again; a reader deciding whether the release affects them wants to know that reading a contract index element inherited from a concurrently-removed commit was undefined behaviour. Both commit bodies are also empty — for a UB fix in the store, the *why* belongs in the commit message, not only in the PR description.

**Fix:** Open a tracking issue for the dangling-reference UB, link it from the PR and from the entry (`[#NNN]` plus a link definition at the bottom of the file), and reword to state the defect — e.g. "Fix undefined behaviour when reading a contract index element inherited from a commit removed concurrently".

piecrust/src/store/commit.rs
```diff
@@ -227,69 +265,92 @@ impl Commit {
...
+                panic!("commit store mutex should not be poisoned")
+            }
+        }
+    }
+
+    #[test]
+    fn test_no_dangling_commit_index_reference() {
+        let commit_store = Arc::new(Mutex::new(CommitStore::new()));
+        let contract_id = ContractId::from_bytes([1; 32]);
+        let ancestor_hash = Hash::new(b"ancestor");
```
moCello: **NOTE:** The regression test covers only a single-level ancestor resolving to `ElementOwner::Commit`; the `ElementOwner::Main` branch and the multi-hop base-chain walk are never exercised.

**Context:** `ElementOwner::Main` is reached whenever the base hash is not a live commit — exactly what happens once commits are finalized away, the lifecycle this PR is reasoning about.

**Fix:** Extend the test to a two-level chain (descendant → intermediate → ancestor) and to a base hash that resolves through `main_index`, so all three resolution paths run under Miri.

piecrust/src/store/commit.rs
```diff
@@ -227,69 +265,92 @@ impl Commit {
...
+    pub fn deep_index_get<'a>(
         contract_id: ContractId,
-        commit_store: Option<Arc<Mutex<CommitStore>>>,
+        commit_store: Option<&'a Arc<Mutex<CommitStore>>>,
         base: Option<Hash>,
-    ) -> Option<*const ContractIndexElement> {
-        if let Some(e) = index.get(&contract_id) {
-            return Some(e);
-        }
+    ) -> Option<ContractIndexElementRef<'a>> {
```
moCello: **NOTE:** With `deep_index_get_mut` gone, `Hulk` is a stateless unit struct wrapping one associated function with one caller — a named boundary that adds nothing.

**Fix:** Inline the body into `Commit::index_get` and delete `Hulk`. The `commit_store` and `base` arguments are just `self`'s own fields.

piecrust/src/store/commit.rs
```diff
@@ -203,17 +206,26 @@ impl Commit {
...
+        let commit_store = commit_store.lock().unwrap();
+        loop {
+            let (maybe_element, commit_base) =
+                commit_store.get_element_and_base(&base, contract_id);
+            if let Some(element) = maybe_element {
+                return Some(f(element));
+            }
+            base = commit_base?;
+        }
     }
```
moCello: **WARNING:** `f` runs while the process-wide `CommitStore` mutex is held (base-chain branch only), with nothing in the signature or a doc comment saying so — a closure that re-enters the store deadlocks the VM.

**Context:** `std::sync::Mutex` is not reentrant, and this same lock is taken by `sync_loop`, `remove_commit`, and every commit/finalize path — so any store access from inside `f` self-deadlocks or blocks the commit pipeline. All four in-tree callers pass pure closures, so nothing hangs today; the hazard is the next caller, invisible without reading the body.

**Fix:** Add a one-line doc comment on `with_index_element`: `f` may run while the commit-store mutex is held and must not access the commit store.

PR #487, opened 2026-06-30
piecrust: reject contract merkle position collisions

d-sonuga:
Add missing check to reject deployment of a contract that maps to an unavailable Merkle position.

### DIFF
--- a/piecrust/src/store/session.rs
+++ b/piecrust/src/store/session.rs
@@ -20,7 +20,7 @@ use crate::store::baseinfo::BaseInfo;
 use crate::store::commit::Commit;
 use crate::store::commit_store::CommitStore;
 use crate::store::hasher::Hash;
-use crate::store::tree::PageOpening;
+use crate::store::tree::{PageOpening, position_from_contract};
 use crate::store::{
     BASE_FILE, BYTECODE_DIR, Bytecode, Call, ELEMENT_FILE, MAIN_DIR,
     MEMORY_DIR, METADATA_EXTENSION, Memory, Metadata, Module,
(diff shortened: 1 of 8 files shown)


piecrust/src/store/session.rs
```diff
@@ -372,6 +372,86 @@ impl ContractSession {
...
+        // Replace is used by migration: the target ID may already exist and, in
+        // that normal case, reusing its own base Merkle position is correct.
+        // The pending replacement ID is about to be removed and re-keyed to the
+        // target ID, so it must not count as a collision with the target.
+        // Reject only when the position belongs to a different pending/base ID.
+        let pending_collision = self.pending_contract_position_occupied(
+            contract_id,
+            pos,
+            Some(replacement_contract),
+        );
```
moCello, line 428: **NOTE:** The `pending_collision` branch of the replace guard is untested — the three new tests cover deploy-pending, deploy-base, and replace-base, but not this path, in an elevated-care state-storage area.

**Context:** The gap is a same-session pending contract — other than target and replacement — folding to the target position. Reachable via migrate after a third contract is deployed at that position earlier in the session.

**Fix:** Add a migrate test that deploys a third contract at the target position in the same session before the migrate, asserting replace is rejected via `pending_collision`.

moCello (approved):
APPROVE — Correct, well-tested fix. A deploy whose id folds onto an already-occupied Merkle position used to silently overwrite the existing leaf; this rejects it, checking both pending session state and the committed base tree (in-memory and disk-reconstruction paths both covered). Both prior-round comments are addressed.

One thing for rollout, not a blocker here: a deploy that collides with an existing contract's Merkle position is now rejected where it used to be accepted (non-colliding deploys are unaffected), so on a live network mixed-version nodes would diverge on such a deploy. When it's pulled into rusk, gate it behind a coordinated activation height. The one note below is minor.

Neotamandua (changes requested):
The stringly-typed domain errors are being passed through io::Error::other and then at other places are wrapped as Error::PersistenceError.

The collision case is a deterministic VM/state invariant, not really an I/O persistence failure. The change creates a collision error as io::Error::other(format!(...)), returns it directly from the deploy-side helper, but in other places wraps the same kind of error into Error::PersistenceError from a helper.

I would like to see that we add a typed error variant ourselves, for example:
```rust
pub enum Error {
    // ...

    #[error("Contract already exists: {0}")]
    ContractAlreadyExists(ContractId),

    #[error("Contract '{contract_id}' collides at Merkle position {pos}")]
    ContractPositionCollision {
        contract_id: ContractId,
        pos: u64,
    },
}
```

If I see it correctly, for the “not found” case, we could reuse the existing Error::ContractDoesNotExist(ContractId) variant.

piecrust/tests/persistence.rs
```diff
@@ -316,3 +316,46 @@ fn migration_self_id_remains_same() -> Result<(), Error> {
...
+        .migrate(
+            id_a,
+            contract_bytecode!("box"),
+            ContractData::builder().owner(OWNER),
+            LIMIT,
+            |_, _| Ok(()),
+        )
+        .expect_err("migration to colliding contract id should be rejected");
+
+    assert!(matches!(err, Error::PersistenceError(_)));
```
Neotamandua: The tests assert only Error::PersistenceError(_). That is weakly typed. An unrelated persistence failure could pass here too. 

Once the error is typed, we can assert the exact variant and fields, e.g. ContractPositionCollision { contract_id, pos } instead of checking a broader wrapper

piecrust/src/store/session.rs
d-sonuga: Added a test for this.

PR #488, opened 2026-07-01
piecrust: remove unused indexmap dependency

Neotamandua:


d-sonuga (approved):

moCello (approved):
APPROVE — Dependency confirmed unused across all source and manifests; the removed version pin's edition-2024 rationale is already obsolete (workspace is edition 2024, MSRV 1.94), and Cargo.lock is untracked so no lock desync. Clean single-concern removal with a correct changelog entry. Nothing blocks.

PR #490, opened 2026-07-01
piecrust: update chain WASM config

Neotamandua:


piecrust/src/vm.rs
```diff
@@ -60,6 +60,22 @@ fn config() -> Config {
     // Support 64-bit memories
     config.wasm_memory64(true);
 
+    // Keep SIMD but force determinism.
+    config.wasm_simd(true);
+    // Reject relaxed SIMD as it can yield host-dependent results.
+    config.wasm_relaxed_simd(false);
+    // Only allow deterministic SIMD operations, this
+    // is only relevant if relaxed SIMD is enabled.
+    config.relaxed_simd_deterministic(true);
```
herr-seppia, line 69: Why do we need to set it if relaxed_simd is disabled?

herr-seppia (approved):
LGTM

Ps: does this change current contracts behavior in any way?

piecrust/src/vm.rs
Neotamandua: > Why do we need to set it if relaxed_simd is disabled?

explicitness of config & for future determinism, if we ever intend to enable relaxed simd

